PR #176, opened 2015-10-05
Adding ability to directly supply index paths

godotgildor:
This pull request addresses the following main points, with a common theme of allowing more flexibility in the paths of supporting files (for instance .bai companions to .bam files).

1. It updates the IGV session writer to save the index file path into the session file if there is an index file known.  The native IGV session writer already saves the coverage file path if known, and has a field to save the index file, but at present the writer did not actually write the index path.
2. This pull request makes it easier to provide the paths of index files which may differ from the paths of the files they support.  For instance, some users may keep their .bam files in one directory, and supporting files like .bai or .tdf in separate directories.  To make it easier to explicitly set these paths this pull request includes:
  * A new menu item allowing for users to give the path to a bam file and a separate path to the .bai file that belongs to that bam.
  * Introduces a new session file based heavily on the UCSC session file.  The main difference is this session file has fields to supply index and coverage file paths, and just as the UCSC session file allows for users to simply give a list of paths for files, one path per line, this session file allows users to provide paths along with their index file paths and coverage file paths.  For instance, let's say a user wants to load in a couple of bam files and a vcf.gz file into IGV.  With this "index aware" session file, they can simply write:
  ```
  /path/to/foo.bam /different/path/to/foo.bam.bai
  /path/to/bar.bam /different/path/to/bar.bam.bai /yet/another/path/to/bar.bam.tdf
  /another/path/to/bob.vcf.gz /last/path/to/bob.vcf.gz.tbi
  ```
This makes it very easy to import a set of reads/variants with their corresponding index files and coverage files.  Then with the update in the IGV session writer mentioned previously, saving this as an IGV session will now contain all information to reload this data easily.
3. In the IGV source, there is existing special case logic for files from GoogleCloud or Ga4ghAPIHelper so that it will not attempt to find helper files (like coverarage files) by adding .tdf onto the path and checking for a file to exist.  This pull request makes this more generalizable by providing a check box item in preferences allowing users to selectively disable this "helper function auto-discovery".  This enables additional environments where such auto-discovery is not appropriate.

### DIFF
--- a/src/org/broad/igv/PreferenceManager.java
+++ b/src/org/broad/igv/PreferenceManager.java
@@ -209,6 +209,7 @@ public class PreferenceManager implements PropertyManager {
     public static final String USE_PROBE_MAPPING_FILE = "USE_PROBE_MAPPING_FILE";
 
     public static final String SEARCH_ZOOM = "SEARCH_ZOOM";
+    public static final String BYPASS_FILE_AUTO_DISCOVERY = "BYPASS_FILE_AUTO_DISCOVERY";
     public static final String NORMALIZE_COVERAGE = "NORMALIZE_COVERAGE";
     public static final String SHOW_EXPAND_ICON = "SHOW_EXPAND_ICON";
     public static final String SHOW_DEFAULT_TRACK_ATTRIBUTES = "SHOW_DEFAULT_TRACK_ATTRIBUTES";
@@ -1060,6 +1061,7 @@ public class PreferenceManager implements PropertyManager {
         defaultValues.put(SAM_COMPLETE_READS_ONLY, "false");
         defaultValues.put(SAM_SHOW_ALL_BASES, "false");
 
+        defaultValues.put(BYPASS_FILE_AUTO_DISCOVERY, "false");
         defaultValues.put(NORMALIZE_COVERAGE, "false");
 
         defaultValues.put(SHOW_GENOME_SERVER_WARNING, "true");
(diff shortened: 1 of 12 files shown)


jrobinso:
Hi,  apologies for the delayed response,  I've been traveling this month and otherwise engaged and haven't had time to review this.    I think (1) and (3) could be pulled as is or with minimal changes.  RE (2),  I have some comments.    
* I haven't pulled to see what the new menu item for index looks like.   In general I think this will be fairly rare, and the File menu is already quite busy.   However I don't rule it out,  I can see the need.   One thought was to modify the error dialog that popups up when the index is not found to allow you to enter it there.   Another is to add extra menu items optionally, triggered by user preference.  
* IGV already supports reading some UCSC session files,   although it doesn't write them.  If possible I would prefer additions be supersets of the UCSC session file,  so that files created could be used for IGV and UCSC.   I don't know if this is possible, or if UCSC would choke on the extra attributes.   Also,  I should mention that there will be a new session format based on JSON rather than XML at some point.  This will create sessions that can be used with igv.js.   The XML session will be supported indefinitely as legacy files,  but eventually I want to transfer to JSON.

Again sorry for the delay,  these are very good changes will be welcome I'm just swamped right now.

jrobinso:
I had to merge this in a local branch, then merge the branch to master and push.  Could you pull and verify that all your changes made it through the process?  My "git foo" is not what it should be.   I will likely merge the 2 "Load from url" menu items into one before the next release, perhaps with a multi-field dialog to enter the 2 urls.

godotgildor:
For some reason I missed the notifications about this pull request.  Thanks for accepting the changes!

Issue #183, opened 2016-01-08
Feature request:


yfarjoun:
in the knowledge-base plugin (for the GATK NA12878 knowledgeBase) one cannot add a review to a multiallelic site (the dialogue box doesn't accept). It would be good if in this case the plugin would  subcontext the variant to genotype that was clicked when attempting to submit the review.

I am aware that this feature is only aimed ad the GATK team at broad, and so we will implement it. I just wanted to document the feature here.

jrobinso:
OK Yossi, let me know if you need anything from me.

ldgauthier:
I'm seeing some test failures when I run on the master branch (namely RunPluginTest, CoverageTrackTest, TrackLoaderTestHeaded, and PreferencesEditorTest).  My Java version is 1.7.0_51.  Is that a problem?

jrobinso:
The java version is fine.  Could you post the test results?

ldgauthier:
Here are some excerpts from the results:
  ```
    <testcase classname="org.broad.igv.sam.CoverageTrackTest" name="testAfterBAMRemoved" time="0.402">
          <failure message="expected:&lt;4&gt; but was:&lt;5&gt;" type="junit.framework.AssertionFailedError">junit.framework.AssertionFailedError: expected:&lt;4&gt; but was:&lt;5&gt;
        at org.broad.igv.sam.CoverageTrackTest.testAfterBAMRemoved(CoverageTrackTest.java:79)

     <testcase classname="org.broad.igv.session.IGVSessionReaderTestHeaded" name="testLoadCombinedDataSourceSession" time="1.359">
          <error message="java.lang.IllegalStateException: Could not find track with ID test/data/seg/Broad.080528.subtypes.seg.gz_TRIBE_p_TCGAaffx_B1_2_GBM_Nsp_GenomeWideSNP_6_E11_155884" type="java.lang.RuntimeException">java.lang.RuntimeException: java.lang.IllegalStateException: Could not find track with ID test/data/seg/Broad.080528.subtypes.seg.gz_TRIBE_p_TCGAaffx_B1_2_GBM_Nsp_GenomeWideSNP_6_E11_155884
        at org.broad.igv.ui.IGV.restoreSessionSynchronous(IGV.java:1463)
        at org.broad.igv.ui.IGV$8.run(IGV.java:1397)
        at org.broad.igv.util.LongRunningTask.submit(LongRunningTask.java:57)
[...]

PR #184, opened 2016-01-15
In VariantReviewDialog, subset multiallelic VC to just the sample rel…

ldgauthier:
…evant to the dialog box so that we can submit a review

Expand dialog box to show entire stop position field; change "mut" field to "alleles"; separate alleles with comma to prevent confusion with genotype

Implements #183


ldgauthier:
Tests pass locally

PR #246, opened 2016-05-31
Support long reads in "BLAT a sequence"

### LINKED ISSUE #245: "BLAT a sequence" fails for long reads
The "BLAT a sequence" operation fails for long reads.  The problem seems to be that the query sequence is passed to BLAT as a URL parameter, and the URL length limit is exceeded at ~8kb.  This causes problems for most PacBio reads, which are regularly > 10kb.

amwenger:
The ﻿"BLAT a sequence" operation fails for long reads because the query sequence is passed to BLAT as a URL parameter and the URL length limit is exceeded at ~8kb.  This adds a test for the long read situation and provides a fix by passing the query sequence via POST instead of as a URL parameter.

Closes #245.


jrobinso:
Thanks for the fix!

PR #266, opened 2016-06-29
Minor bugfix: incorrect null check safeguard

amwenger:
Minor bugfix submitted as a unit to keep pull requests atomic... more substantive work on the way.


jrobinso:
Thanks for the pull request,  I note its in IonTorrent related code.  Apparently you use that?  I ask because no one maintains it, and we are considering dropping it as Ion Torrent has their own fork of IGV.

jrobinso:
Ignore last comment (deleted),  the flow signal section change was just white space deletion.

PR #272, opened 2016-07-06
Improve alignment details tooltip

amwenger:
Improve the alignment details tooltip information:

1. Abbreviate long CIGAR strings in the middle instead of at the end.  This is useful with third generation long read sequence data, which can have very long CIGAR strings.  The beginning and the end of the CIGAR, which show clipping information, are particularly informative.
2. Add an explicit line to show soft/hard clipping information.
3. Add the `SAM_HIDDEN_TAGS` option to suppress certain tags from the tooltip (default to "MD,SA").  Some tags, like `MD` and `SA` are verbose and not readily understandable by a human reader.  The tags can consume a huge amount of screen space with long read data.
4. Add the read length.
5. List the full alignment reference span, not just the alignment start.
6. Consolidate mapping information into a single line.
7. Consolidate basepair information into a single line.

An example before & after is:

alignment details modifications

### DIFF
--- a/src/org/broad/igv/PreferenceManager.java
+++ b/src/org/broad/igv/PreferenceManager.java
@@ -107,6 +107,7 @@ public class PreferenceManager implements PropertyManager {
     public static final String SAM_FILTER_SECONDARY_ALIGNMENTS = "SAM.FILTER_SECONDARY_ALIGNMENTS";
     public static final String SAM_FILTER_SUPPLEMENTARY_ALIGNMENTS = "SAM.FILTER_SUPPLEMENTARY_ALIGNMENTS";
     public static final String SAM_FILTER_URL = "SAM.FILTER_URL";
+    public static final String SAM_HIDDEN_TAGS = "SAM.HIDDEN_TAGS";
     public static final String SAM_MAX_VISIBLE_RANGE = "SAM.MAX_VISIBLE_RANGE";
     public static final String SAM_SHOW_DUPLICATES = "SAM.SHOW_DUPLICATES";
     public static final String SAM_SHOW_SOFT_CLIPPED = "SAM.SHOW_SOFT_CLIPPED";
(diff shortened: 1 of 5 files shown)


jrobinso:
Nice, thanks for the PR

jrobinso:
BTW,  the next major release (3.0) will have a lot of improvements and new features specifically for pacbio and other long reads, as well as support for 10X tagged bams.   Is this of interest?  If so perhaps we should open an issue to discuss proposed changes.

amwenger:
Better PacBio and 10X support will be very useful!  I have a number of
local modifications (amwenger/igv fork) to improve IGV support for long
reads and have been trying to package them to contribute to the general IGV
project.

I would certainly participate in a feature discussion.  Could you open the
issue to get the ball rolling with your plans?

jrobinso:
Sure,  I'll try to do it today or tomorrow,  I need to review what I've 
already done and write something down.   I worked on pacbio a year ago,  
implementing a number of features that they (pacbio) requested but never 
got input or feedback on them and that sort of stagnated.   I'm 
currently working on 10X.   First step is to create "Loupe" type linked 
read views using their barcodes.

I'll open an issue for each,  ping me again on Monday if they aren't 
open or just do it yourself.

amwenger:
Sounds great.  As of two months ago, I am a research scientist at PacBio
(coming from academics and clinical work with mostly Illumina data).  I
will provide input and feedback on PacBio-related IGV features and am happy
to do some legwork too.

I will open the "long read support" ticket and put my ideas down (probably
tomorrow).

jrobinso:
OK, all the previous work was with pacbio engineers at their specific 
direction,  and then they just disappeared so-to-speak. Happy to pick 
this up again,  but a full person-mothh of work has already been done 
and we should pick up there.   It will take me a day or two to get back 
up to speed.   If you ask around there you might find the people 
involved in the last round, I don't want to post emails in a public 
forum here.

jrobinso:
Issue opened for 10X discussion: https://github.com/igvteam/igv/issues/275

jrobinso:
After nearly going nuts trying to find my SA tags I realized they were hidden by default.   After some consideration I'm not sure this is a good idea.  Perhaps for pacbio they are verbose and hard to interpret,  but for other platforms perhaps not?   For the time being I have reverted to no tags hidden by default, this is under the users control of course.   Perhaps we can have defaults by platform, or have a better UI to display SA tags in an intelligible way.

amwenger:
Sorry for the confusing default!  I agree that defaulting to no hidden tags is the best platform-independent decision.

The SA tag is hard to read for PacBio reads because it includes the CIGAR string for other alignments.  PacBio CIGAR strings can be incredibly long (because of single pass indel errors).  IGV has logic to abbreviate the CIGAR of an alignment, but not the CIGARs embedded within SA tags.  I will think about how to better display it, but for now, I agree with your change.  Thanks!

Issue #277, opened 2016-07-12
Improved support for PacBio reads


amwenger:
IGV is a very useful viewer for PacBio long read data, but it could be even better with a few modifications.  Ideas to improve support for PacBio data are:
1. **Add a "quick consensus" mode** that only shows mismatches at positions with a consistent alternative variant.  The coverage track has logic that uses an allele threshold to determine whether to show variation at a position.  Add a "quick consensus" option and apply the same logic to mismatches within a read.  This would greatly hide the "random errors" in long reads and make it possible to see haplotype structure by eye.  It would look something like this:
   ![quick-consensus](https://cloud.githubusercontent.com/assets/7155109/16758576/3c93dd42-47c7-11e6-9b58-b23794c0b0f6.png)
2. **Hide small indels** at low zoom.  Individual PacBio reads have random small deletions errors that show as a series of black dots at low zoom (see picture below).  Hide deletions that occupy <3px of at the current zoom.
   <img width="631" alt="many-deletions" src="https://cloud.githubusercontent.com/assets/7155109/16758579/41e5b414-47c7-11e6-98ac-0ba1819c1960.png">
3. **Label the size of large insertions and deletions**.  Modify the "Flag insertions large than N bases" option to be "Label insertions and deletions larger than N bases".  "Large" insertions and deletions would show at all zoom levels.  One idea for how to show the variant size is to use a filled upward pointing trapezoid for insertions and a hollow download pointing trapezoid for deletions:
   ![insertion-size](https://cloud.githubusercontent.com/assets/7155109/16758582/45b03312-47c7-11e6-9470-dcfe8f6ee86c.png)
   ![deletion-size](https://cloud.githubusercontent.com/assets/7155109/16758585/47bc7a62-47c7-11e6-874f-b25df68cf3be.png)
4. **Show clipping information at the end of reads**.  Read clipping can indicate the presence of a structural variant.  Show the number of clipped bases in a "cap" at the end of reads:
   ![cliipping-cap](https://cloud.githubusercontent.com/assets/7155109/16758588/4c2b6536-47c7-11e6-89b6-93328386527d.png)
5. **Show variation at low zoom** to enable viewing haplotype structure.  I would prefer to see SNVs and large indels even when zoomed out to 100kb+.
[...]

PR #285, opened 2016-07-18
Add "quick consensus" mode to show only consistent mismatches

amwenger:
Add a "quick consensus" mode feature controlled by a setting that limits
the display of basepair mismatches to positions with a consistent
alternative base.  This is especially useful with third generation
sequencers that have a high single-read mismatch error rate.  The
consensus mode uses the same logic as the coverage track to determine
which mismatches to show.

Addresses point 1 in #277

### DIFF
--- a/src/org/broad/igv/PreferenceManager.java
+++ b/src/org/broad/igv/PreferenceManager.java
@@ -110,6 +110,7 @@ public class PreferenceManager implements PropertyManager {
     public static final String SAM_HIDDEN_TAGS = "SAM.HIDDEN_TAGS";
     public static final String SAM_MAX_VISIBLE_RANGE = "SAM.MAX_VISIBLE_RANGE";
     public static final String SAM_SHOW_DUPLICATES = "SAM.SHOW_DUPLICATES";
+    public static final String SAM_QUICK_CONSENSUS_MODE = "SAM.QUICK_CONSENSUS_MODE";
     public static final String SAM_SHOW_SOFT_CLIPPED = "SAM.SHOW_SOFT_CLIPPED";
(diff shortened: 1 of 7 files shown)


jrobinso:
Hi,  could you add a few instructions here for invoking this?   Also, without seeing it in action its difficult to guess why BedRenderer and FeatureRenderer would be impacted.   Does this mode add new tracks?

amwenger:
> could you add a few instructions here for invoking this?

An example BAM file is available [here] (https://www.dropbox.com/s/vo368hpwb0qri1n/hg38.foxp2test.bam?dl=0).

To see the "quick consensus" mode in action, load the example BAM and navigate to hg38 chr7:114,317,353-114,321,356.  In standard mode, the locus looks like:
screen shot 2016-07-18 at 3 45 44 pm

To turn on "quick consensus" mode, go to View > Preferences > Alignments and check "Quick consensus mode":
screen shot 2016-07-18 at 3 46 06 pm

The same "coverage allele-fraction threshold" that is used to determine whether to draw a frequency histogram in the coverage track is used to determine whether to draw a mismatch in an alignment.  With consensus mode and a threshold of 0.25, the locus looks like:
screen shot 2016-07-18 at 3 46 30 pm

Change the threshold to 0.75 to see no variants.

> Also, without seeing it in action its difficult to guess why BedRenderer and FeatureRenderer would be impacted. Does this mode add new tracks?

This mode does not add any new tracks.  To determine whether to show variation at a position, the logic needs the `alignmentCounts`.  I get the counts in `AlignmentTrack.java` and pass them to the renderer, which is of type `FeatureRenderer`.  So, I had to change the function signature of `drawAlignment()` in `FeatureRenderer` and its descendant class, `BedRenderer`.

PR #295, opened 2016-08-02
Hide small indels at low zoom

amwenger:
Individual PacBio reads have random small deletion and insertion errors that obscure the important details in read alignments.  This pull request matches the resolution of alignment information to the zoom level: small insertions and deletions are hidden at low zoom; large indels are visible at all zooms.  The logic calculates the pixel width of the indel at the current zoom and only renders indels of at least a minimum pixel size.

Implementation of the indel hiding required modifying the SAM `drawAligment()` method.  Rather than walking through alignment blocks and drawing the blocks one by one, the new logic walks through the alignment gaps and draws the flanking blocks.  In my opinion, the new approach is a bit easier to understand, and it is certainly easier for skipping small gaps.

Low zoom without hiding small indels (BEFORE):
orig

Low zoom with hiding small indels (AFTER):
smoothed

Medium zoom without hiding small indels (BEFORE):
orig-indel-zoom

Medium zoom with hiding small indels (AFTER):
smooth-indel-zoom

This addresses suggestion 2 in #277.

### DIFF
--- a/src/org/broad/igv/sam/AlignmentRenderer.java
+++ b/src/org/broad/igv/sam/AlignmentRenderer.java
@@ -80,6 +80,9 @@ public class AlignmentRenderer implements FeatureRenderer {
 
     private static Stroke thickStroke = new BasicStroke(2.0f);
 
+    // Minimum pixel width to render a small indel
+    private final int MIN_INDEL_PX_WIDTH = 3;
+
     // Bisulfite constants
     private final Color bisulfiteColorFw1 = new Color(195, 195, 195);
     private final Color bisulfiteColorRev1 = new Color(195, 210, 195);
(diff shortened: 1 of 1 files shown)


jrobinso:
This is a good start at this problem.   I think we will have to make it configurable though.  For non-pacbio platforms you often do want to see indels when zoomed out.    Another approach to the pacbio indel problem,  suggested by Jason in the last round, was to simply ignore indels < some threshold width in bp.   These are not mutually exclusive options,  and obviously have different effects when zoomed in.

amwenger:
Having the behavior be configurable is a good idea.

It's harder to implement, but perhaps the best behavior for PacBio data would be similar logic as for the basepair mismatch "quick consensus".  If an indel is consistent across reads, then show it.  If not, then suppress it.

jrobinso:
This is now configurable.   Also, indels can be hidden based on a pixel or base pair threshold.

amwenger:
Very nice.  I suggest a few tweaks in #309.

It will be good to have some user feedback over whether the min pixel width is understandable.  Your min basepair option is more explicit and probably better than my initial pixel approach.

PR #304, opened 2016-08-10
V2.3.x

justinthezhu:
Could you set the data range of the BAM coverage to be the same for multiple BAM files within one IGV session from a command accessible through an external port?

Additionally, could you create a way to access IGVtools from an external port and save an IGV session through an external port?

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CompilerConfiguration">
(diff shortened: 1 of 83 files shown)


jrobinso:
Hi, sorry but I'm not able to review a pull request with 79 changed files.   Could you redo this against a fresh pull from the master trunk,  and reduce the scope if possible?   Its not clear why so many files need changed for this request.

jrobinso:
Also note that the branch has many conflicts,  these should be resolved before submitting the pull request.

PR #306, opened 2016-08-12
Label the size of large indels

amwenger:
Transform the "flag large insertions" option into an option to label the size of large insertions and deletions.  For large events, the size of the event (in basepairs) is show in a label.  This is particularly useful to differentiate insertion events of different scales and to see the exact basepair size of deletions.

For deletions larger than the threshold, label the basepair size if the label fits over the gap line.  For insertions larger than the threshold, expand the purple "cursor" used for small insertions to be a label with the basepair size of the insertion.  Limit the width of the expanded cursor to the number of pixels that the inserted basepairs would use.

Example:
screen shot 2016-08-11 at 4 58 57 pm


amwenger:
I think this solution is pretty useful, but the exact form does merit some further discussion.  IGV now has an extended concept of "short" and "long" insertions and deletions:

* Short insertions show as a thin purple cursor, independent of the length of the insertion.  This is the same as before this pull request.
* Long insertions shows as a thick purple cursor, where the thickness is proportional to the size of the insertion.  The basepair size of the insertion is added as a label, if the label fits.  Before this pull request, long insertions showed as a thick red cursor, independent of the length of the insertion.
* Short deletions show as a gap line proportional to the length of the deletion.  This is the same as before this pull request.
* Long deletions also show as a gap line proportional to the length of the deletion.  The basepair size of the deletion is added as a label, if the label fits.

I can see an argument to:
1. Remove the concepts of short v long.  Treat all insertions and deletions as "long" ones are handled in the pull request (though perhaps do not label the basepair size of 1bp events).
2. Treat all insertions and deletions as in the pull request but use the concept of "long" to color the long events differently to make them stand out more.
3. Keep the concept of short v long insertion but treat all deletions as handled in the pull request (again except for not labeling 1bp events).

Thoughts?

jrobinso:
Let's put it out there and get some feedback.

PR #333, opened 2016-11-16
Further RNA structure file format support

shapemapper:
This PR is mainly aimed at simplifying loading RNA base pair info or chemical probing data from other programs.

Added semi-automated file conversion on load for
.ct (connectivity table format),
.db/.dbn (dot-bracket notation),
.dp (pairing probability)
.shape/.map (chemical probing reactivity profile)

Upon attempting to open a file with one of these extensions, a dialog appears allowing the user to select the applicable chromosome and specify a new strand and start coordinate if needed (this information is typically not present in any of these input file formats, since they focus on single RNAs). A new file is written (.bp for base pair structure info or .wig for reactivity profile), then that file is loaded.

It's not perfect, but I think this will simplify things so people don't have to run separate scripts.

- Steve Busan

### DIFF
--- a/src/org/broad/igv/feature/BasePairFileUtils.java
+++ b/src/org/broad/igv/feature/BasePairFileUtils.java
@@ -28,8 +28,6 @@ package org.broad.igv.feature;
 
 import org.broad.igv.feature.basepair.BasePairFeature;
 import org.broad.igv.Globals;
-import org.broad.igv.util.ParsingUtils;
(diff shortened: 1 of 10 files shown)


shapemapper:
Here are some example input files. The E. coli ones can use the U00096.2 genome. For HIV I've included a fasta file to create a genome from.

[example_files_to_convert.zip](https://github.com/igvteam/igv/files/593468/example_files_to_convert.zip)

PR #342, opened 2016-12-16
Amw highlight read clipping

amwenger:
Highlight read clipping, which often indicates a structural variant, with a pink
mark at the edges of read alignment that are clipped by more than a specified
number of basepairs.

As an example, here is the same data on the left and right with clipping >5bp
highlighted on the right.  The clipped reads mark 4 structural variant breakpoints.
show-read-clipping

Address idea 4 from #277.


olechnwin:
I downloaded IGV snapshot 03/20/2018, I cannot find this feature. Is it in? How do I use this?

amwenger:
Yes, the feature is in the snapshot.  To activate it: `View > Preferences > Alignments` then check "Flag clipping > flag clipping threshold" and set the threshold in the box below.  Depending on your data type, you might also do `View > Preferences > Third Gen`.

jrobinso:
That's an experimental feature,  the idea is you can have a specific set of preferences for 3rd gen alignments.   However I can see the potential confusion here.

PR #348, opened 2016-12-22
Fix and simplify "group alignments by" feature

amwenger:
This fixes a bug in the "group alignments by" feature and simplifies some of the related preference logic.

Previously, when a grouping option was selected, it was applied to the currently loaded alignment tracks.  When a new alignment track was loaded, it was not grouped.  Further, when a preference change caused a reload (e.g. "Filter secondary reads" change), the grouping was lost.  This slightly consolidates the grouping logic and ensures that the grouping is persistent across loads and reloads.

A version of the problem remains in the sort logic.  If the reads are sorted and a preference is changed, then sort order is lost.

This fix also provides an occasion to reduce data reloads upon preference changes.  Some preference changes (such as "quick consensus mode") require a redraw but not a reload.


jrobinso:
Thanks for sorting through that, especially the refreshes.  Its been on my todo list for a while.

PR #354, opened 2017-01-06
Add BAM alignment track "collapsed" mode

amwenger:
Introduce a "collapsed" mode for BAM alignment tracks that is intermediate between expanded and squished.  In collapsed mode, the individuals reads are drawn at about half the height as in expanded mode.  This also revealed some opportunities to improve the rendering logic to be more robust.

The collapsed mode looks as follows:
ecs

ecs-zoom

### DIFF
--- a/src/org/broad/igv/sam/AlignmentRenderer.java
+++ b/src/org/broad/igv/sam/AlignmentRenderer.java
@@ -678,7 +678,7 @@ public class AlignmentRenderer implements FeatureRenderer {
 
         // Define a graphics context for indel labels.
         Graphics2D largeIndelGraphics = context.getGraphics2D("INDEL_LABEL");
-        largeIndelGraphics.setFont(FontManager.getFont(Font.BOLD, h-1));
+        largeIndelGraphics.setFont(FontManager.getFont(Font.BOLD, h-2));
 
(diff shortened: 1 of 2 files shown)


jrobinso:
Aaron, this would be counter-intuitive wrt how other tracks work.  Collapsed is always smaller than squished.   If we need an intermediate mode between squished and expanded I would prefer it have another name ("Compressed").    This could be in the menu only,  the internal mode value could still be "collapsed".

jrobinso:
As there are other bug fixes / cleanup here I will accept it, then change the menu.   BTW my thought for "collapsed" has been that alignments completely disappear,  leaving only coverage,  but there is already a way to do that.

amwenger:
@jrobinso Thank you for accepting the request and explaining the relative ordering of "expanded", "squished", and "collapsed."   It makes sense that the coverage track alone is most consistent with what collapsed means in other contexts like BED files.  "Compressed" is a good label for the new mode; "thin" or "narrow" could work too.  If you prefer to drop the mode for consistency, it will be easy enough to just leave the bug fixes.  I can help with that if you make the call.

jrobinso:
I accepted the pull request we just need to remember to change the label in
the menu.   This is not a big deal.   I'm going to stay focused on this
insertion thing for now.

PR #382, opened 2017-03-23
rename files to match public class names for case sensitive system

jstjohn:
I was having compile errors on OSX that looked like the following: 

```
    [javac] Compiling 182 source files to /Users/john/src/igv/testtmp
    [javac] /Users/john/src/igv/test/src/org/broad/igv/sam/lite/BgunzipTest.java:12: error: class BGUnzipTest is public, should be declared in a file named BGUnzipTest.java
    [javac] public class BGUnzipTest {
```

Basically the file names matched except the case did not match the class case. Even though builds have obviously worked for other people in the past, this is probably more correct anyways, and would enable builds on more systems?

Thanks!


jrobinso:
You're a brave man building from master at this state.  I should be branching and will soon.   Thanks for the pull request.

PR #389, opened 2017-04-13
Support for generic oauth2 connection.

david-w-mead:
The updates allow for Mayo to use Microsoft's Azure oauth (IDM) service to authenticate using oauth2. The changes do not break the current google oauth functionality, but override certain google oauth functionality based on the setting in the oauth-config.json file. Our oauth-config.json file is below with the client_id and client_secret removed.

{
  "client_id": "",
  "client_secret": "",
  "authorization_endpoint": "https://login.microsoftonline.com/a25fff9c-3f63-4fb2-9a8a-d9bdd0321f9a/oauth2/authorize",
  "token_endpoint": "https://login.microsoftonline.com/a25fff9c-3f63-4fb2-9a8a-d9bdd0321f9a/oauth2/token",
  "hosts": "mforgeweb.mayo.edu",
  "auth_provider": "Mayo Clinic",
  "app_id_uri": "https://mctools.onmicrosoft.com/986e76c6-cf18-427e-b779-f7af5a36de48",
  "scope": "openid",
  "find_string": "bsi",
  "replace_string": "bsi-oauth2"
}


jrobinso:
Is the Jackson annotation library required?  Where is it used, and what does it add?   Where possible I like to avoid additional 3rd party dependencies.

david-w-mead:
Hi Jim,

Good question. This is clearly a lot of overhead for minimal
functionality.  The purpose of java-jwt and it's 4 dependencies, the
jackson jars and commons-codec, is to extract the username from the token.
Microsoft has not implemented the rest call to return the displayName in
their oauth2 implementation and the only place I could find username was
embedded in the token itself.

This is not a big deal to me (really don't need to display user name), so
you can cut out the lines I added to fetchUserProfile and remove the jars.
We just need to handle the exception when this method is called when using
Microsoft oauth.

Thanks,

David

jrobinso:
Hi,  I attempted to merge this PR but ran into other difficulties.   Why was this change made (moved event package)?    

-import org.broad.igv.event.GenomeChangeEvent;
 -import org.broad.igv.event.IGVEventBus;
 -import org.broad.igv.event.IGVEventObserver;
 +import org.broad.igv.ui.event.GenomeChangeEvent;
 +import org.broad.igv.ui.event.IGVEventBus;
 +import org.broad.igv.ui.event.IGVEventObserver;

jrobinso:
OK,  I think I have this correctly merged, and kept the JWT parsing of the access token.   I removed the jackson libraries, they don't appear to be referenced.  Perhaps they are required in some subtle way?

I have no way to test your setup so could you pull and test?   I think the user profile function really needs some work,  your provider apparently returns JWT tokens that can be parsed for the information.   Google does not.    I'm not sure what the right thing to do here is.

david-w-mead:
This was not intentional. My best guess is it must have been some some
automated eclipse refactoring that I didn't notice.

david-w-mead:
The jackson libraries are used by the java-jwt library.

PR #415, opened 2017-05-23
Initialize OAuthUtils early in the execution so the oauth provider is…

david-w-mead:
… correct.

Removed functionality to show the logged on username when connecting to Microsoft oauth server because Microsoft does not support the functionality.


david-w-mead:
Made a few small changes after testing. I've removed the jars I added before. The only thing that is different is the username won't be displayed when connecting to Microsoft Azure Oauth provider. I added some static code so Mayo Clinic is displayed instead of Google in the Oauth menu when using the Mayo/Microsoft Oauth provider.

PR #417, opened 2017-05-24
Fix #416: black images from batch snapshot

### LINKED ISSUE #416: Batch "snapshot" produces black for the data window
The "snapshot" command in IGV batch scripts is producing all black in the data window.  File > Save Image works fine.  The behavior is the same for me for IGV 2.3, 3.0, and master on macOS Sierra.

Batch script

    genome hg19
    goto chr6:43136920-43151244
    snapshot 1.png

Batch script "snapshot"

File > Save Image
filesave

amwenger:
Fix issue #416 where the batch "snapshot" command produces a fully black image.

This modification does correct the problem for me, but I am not certain that it is the right fix.  This adjusts a code change from Jan 2013 in 9cffc1026274a038aff8f0.

### DIFF
--- a/src/org/broad/igv/batch/CommandExecutor.java
+++ b/src/org/broad/igv/batch/CommandExecutor.java
@@ -874,7 +874,7 @@ public class CommandExecutor {
         }
 
         try {
-            return IGV.getInstance().createSnapshotNonInteractive(target, file, true);
+            return IGV.getInstance().createSnapshotNonInteractive(target, file, false);
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks that's helpful information,  but setting "paintOffscreen" to false has other consequences.  The main one being that you get an exact representation of the window (with scrollbar for example), and cannot set panel heights.    It is a good clue as to the underlying problem.

jrobinso:
Fixed by other means

Issue #361, opened 2017-01-19
Variants being shown sporadically


Lenbok:
Hi, I've been using current master and have an issue with variant track contents sometimes not being displayed. You did say in my other issue that master is in a state of flux, so feel free to say "just use the PRE_INS tag" :-), but thought you might not already know about this.

Here are a couple of screenshots:

![image](https://cloud.githubusercontent.com/assets/282098/22124516/04045004-def5-11e6-86f0-b5a1c1c28c5e.png)

![image](https://cloud.githubusercontent.com/assets/282098/22124537/2406e9c0-def5-11e6-9254-6635903e5898.png)

The variant track is only displayed when the marker lines are with 14 bases to the right of the coordinate of the variant.

If I zoom out, the variant track contents are never displayed:
![image](https://cloud.githubusercontent.com/assets/282098/22124700/c21031e4-def5-11e6-9bd1-45a6f90e26c2.png)

jrobinso:
Thanks I wasn't aware of that.   What are the "marker lines" you are referring to?   ("The variant track is only displayed when the marker lines").   


I appreciate the testing,  but for real work use PRE_INS.

Lenbok:
The marker lines I am referring to are the vertical indicators shown in the alignment track.

:+1:  PRE_INS

jrobinso:
I'm not able to reproduce this using 1KG data (session attached).     Do
you see the problem if you load this session?

Lenbok:
I could not reproduce it with your session so far. I then tried re-loading my session that originally had the problem, and can not consistently trigger it - I have had it happen once more, where things were fine at the most zoomed in level, but missing one level above (like my third screen shot). I will keep an eye on it and will let you know if I work out how to trigger it.

jrobinso:
OK, I'll study the VCF track code and see if I can spot a culprit.

jrobinso:
I can't reproduce this,  there were several checkins that might have fixed it.  I'm closing but reopen if you see it again.

Lenbok:
Not sure how to actually reopen this, but I am still seeing the problem, and have what seems to be a way of reliably reproducing. Looks to involve the remote interface (which is the primary way I drive IGV).

* Start IGV
* Select Human (1kg, b37+decoy)
* Select MT
* Click zoom in button three times
* Load the attached VCF from [igv-test.zip](https://github.com/igvteam/igv/files/815263/igv-test.zip) *using the remote interface*
* Click the zoom out button two or three times
[...]

PR #431, opened 2017-06-08
Fix inability to zoom out when tracks were initially loaded from port/batch

Lenbok:
OK, #361 was driving me bonkers, so I dug into this with the debugger.

During batch loading of a VCF, `DataPanel.paintComponent()` is doing `loadInProgress = true` and then calling `Preloader.load()`. Inside that load method, `loadInProgress` is only set back to false when the features are loaded asynchronously, **not** for the case when `Globals.isBatch()` is true.

Even when loading from batch/port, we need to ensure that `loadInProgress`
is reset to false after loading is completed, otherwise subsequent interactive `DataPanel.paintComponent()` calls never call `Preloader.load()` to fetch the variants we need after e.g. zoom out operations.


jrobinso:
Correct. Good catch

PR #432, opened 2017-06-10
Fix crash when choosing files on Windows 10.

dbrant:
There's a nasty crash happening in the latest version of Windows 10 (Creators Update) and the latest JRE, where creating an instance of JFileChooser causes an internal access violation. The bug is documented [here](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8179014), and contains a workaround that I've provided in this patch.

The only consequence for the user is that the FileChooser will explicitly show file extensions, which may not correspond to the user's global preference, but at least it doesn't crash.

### DIFF
--- a/src/org/broad/igv/ui/Main.java
+++ b/src/org/broad/igv/ui/Main.java
@@ -91,6 +91,14 @@ public class Main {
 
         Runnable runnable = new Runnable() {
             public void run() {
+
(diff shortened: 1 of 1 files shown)


jrobinso:
Thank you for the patch!  The other workaround is to set the L&F to something other than Windows, but I like your solution better.

PR #439, opened 2017-07-10
Allow setting NAME_PANEL_WIDTH to 0

twmccart:
Setting NAME_PANEL_WIDTH to 0 was already how igv handles making the Name Panel non-visible, but it throws an exception if it is set to 0 by the user or the prefs.properties file. This change allows the user to make the Name Panel invisible by default.


jrobinso:
Thanks. Merged.

PR #440, opened 2017-07-10
Created Show_mismatches variable

twmccart:
Created a Preference variable SAM.SHOW_MISMATCHES that can be set to false in 'prefs.properties' to prevent igv from automatically showing mismatched bases in alignments. This is useful for users generating figures with batch scripts. Default functionality should otherwise be the same: mismatched base visibility can be adjusted in the alignment track's menu, and by default igv will still show mismatched bases on start.
I believe that this change should be transparent to the user.


jrobinso:
nicely done

PR #470, opened 2017-10-04
Added option to manually copy+paste OAuth URL

julianhess:
Many Linux desktop environments do not support the Desktop.browse() method.  Added a simple fix to check if this action is supported, and if not, open a dialog box allowing the user to manually copy the OAuth URL into their web browser.

If the browse() method is supported, then the web browser will open automatically, like before.


jrobinso:
Nice, thanks.  BTW who is your oAuth provider?  I only have Google available for testing, I would like input from users of other providers.

julianhess:
My OAuth provider is Google as well.  Sorry I can't be more help there!

PR #487, opened 2017-11-17
Genome position output in ruler mode

Seraff:
Hello. I really suffered without this feature. 

screenshot from 2017-11-17 15-41-16

I also can add enabling this tooltip in the preferences. 
In the first commit, I wanted to move the output to the status bar, but there were strange window blinkings when you move through the genome when the sequence track is shown.


jrobinso:
Thanks for the PR.  I'll get it it soon,  I am covered up temporaritly on another project.   My only suggest might be to format the output string to aid readability (e.g. 148,453).

Seraff:
No problem. Yes, good idea, I've added string formatting.

PR #494, opened 2017-11-29
Automatically connect to oauth server when opening a session, if necessary.

david-w-mead:
Currently, igv fails to open a session file if connection to oauth server has not been done. A workaround is to connect manually, then open this session. This just automates that connection process.


jrobinso:
Hi David,  I created another PR to merge this into a branch so I could do a little testing with Google first.   I'm closing this PR,  will merge from the branch oauth_patch.

jrobinso:
Accepting.  Note that this carries forward the incorrect assumption that all Google Cloud Storage resources require oAuth.   This assumption will be relaxed in the future.

PR #518, opened 2018-03-16
Change CGIAlignmentReader.hasIndex to return true.

yasushi-saito:
If hasIndex()=false, loading alignment records via "Open URL" menu fails always.
It doesn't hurt to change hasIndex() to be true, since the remote server can
still return an error if it can't handle queries.

Background: I'm trying to load reads from a remote BAM-like (but not quite BAM)
file.  I found CGIAlignmentReader to be the easiest hook to achieve that.


jrobinso:
Thanks.  I was unaware anyone was using this.

PR #522, opened 2018-03-28
Remove duplicate SAM_RELOAD_KEYS list

amwenger:
The list of preferences that trigger a reload is duplicated in `prefs/IGVPreferences.java` and `prefs/Constants.java`, which itself is imported in `IGVPreferences.java`.  I think the list in `Constants.java` should be the single, authoritative version.

I noticed this when adding a new preference in my local branch.  I added the preference to SAM_RELOAD_KEYS list in `Constants.java`, but it had no effect since it was being overridden by the duplicate list in `IGVPreferences.java`.  This PR should not impact current behavior and simply removes duplicate logic that currently has to be synchronized.

### DIFF
--- a/src/org/broad/igv/prefs/IGVPreferences.java
+++ b/src/org/broad/igv/prefs/IGVPreferences.java
@@ -889,26 +889,4 @@ public class IGVPreferences {
     }
 
 
-    /**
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks, I merged this manually.

PR #525, opened 2018-03-30
Be robust to invalid SA entries

### LINKED ISSUE #521: SA:Z:-tags seem to block the mouse-over panels
Dear IGV developers,

I am analysing chimeric alignments to determine possible structural variants. In concorance with the SAM specification I have assigned these reads the `SA:Z:`-tag. Before, I was able to see the mouse-over box that provides the read names etc. . In the latests versions of IGV it seems that these boxes are disabled for reads that have their `SA:Z:`-tag is set.
I have searched through the IGV settings but I did not find an option to enable this. I am not sure if this is a bug or a feature, but since I think it is important to allow inspection of chimeric reads I would kindly ask you to share your thoughts on this and ask you whether there is some kind of work-around?

I hope you can help me a bit further. All best,

Youri

amwenger:
The alignment details popover parses supplementary alignment (SA) entries.
Invalid entries that raise an exception during parsing suppress the popover.
This catches those exceptions and warns about invalid entries.

Closes #521

### DIFF
--- a/src/org/broad/igv/sam/SAMAlignment.java
+++ b/src/org/broad/igv/sam/SAMAlignment.java
@@ -651,8 +651,13 @@ public abstract class SAMAlignment implements Alignment {
         buf.append("SupplementaryAlignments");
         String[] records = Globals.semicolonPattern.split(sa);
         for (String rec : records) {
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks

PR #549, opened 2018-06-15
Fix track names and renaming

mvdbeek:
I noticed that in recent versions of IGV I am unable to rename tracks.

After some digging around I found that the track name always falls back
to the (basename of) the path. It looks like the keys of the track
attribute table `attributes` are not always uppercase, so looking up the
attribute value using the `attributeName` seems to work fine in this
instance. I am not 100% sure this is the correct fix, but if it is not I
hope that this can be helpful in tracking down the issue.


mvdbeek:
Cool, thanks for the fix!

PR #563, opened 2018-07-11
Add "group" and "color" by PacBio movie and ZMW

amwenger:
Add "group by" and "color by" PacBio movie and ZMW, which are similar to grouping or coloring by read group or library.

PacBio read names follow the convention "movie/zmw/qStart_qEnd" where a movie is a single run of PacBio machine with a single SMRT Cell, ZMW is a single hole/molecule on the SMRT Cell, and qStart/qEnd are coordinates for a single subread within a ZMW read.

Just as it is useful to group/color by library and read group, it is useful to group/color by PacBio movie and ZMW.  This PR adds those features with movie and ZMW parsed from the read names.


mpauper:
Thanks to @amwenger for making this. This feature would be very useful for anyone working with PacBio reads. In the simplest scenario it allows for quick and easy identification of library prep or sequencing artifacts.

amwenger:
Thank you, @jrobinso.

PR #573, opened 2018-08-14
Fixes for Google requester-pays buckets (userProject) and derived index URL

jwhitney31337:
Two small fixes to get Google bucket URLs and requester-pays buckets to work properly:

1. Moved code to add userProject to URL earlier in HttpUtils.openConnection , so the change takes effect in all code paths
2. Changed BAMReader.getIndexURL to derive the index URL from the URL returned by HttpsUtils.createURL and not the original "gs://" path.


jrobinso:
Hi,  I'm o.k. with (1), but the change to us the gs:// path was deliberate,  why the change back?

PR #590, opened 2018-10-24
Replace all instances of Feature.getChr with getContig

lbergelson:
@jrobinso It sounds like this is causing problems so I've updated all the deprecated calls to getChr.

* Replacing all instances of `Feature.getChr()` with the equivalent `Feature.getContig()`.
* Fixing some place where `getChr` and `getContig` were implemented differently.
  If they differed I resolved the differences by using the `getChr` implementation.
* Removing redundant implementations of `getChr` and allowing the default
implementation to be used.

### DIFF
--- a/src/main/java/org/broad/igv/batch/CommandExecutor.java
+++ b/src/main/java/org/broad/igv/batch/CommandExecutor.java
@@ -759,7 +759,7 @@ public class CommandExecutor {
             if (locus != null) {
                 int start = Math.max(0, locus.getStart() - 1);
                 String desc = param2 != null ? param2 : "";
-                roi = new RegionOfInterest(locus.getChr(), start, locus.getEnd(), desc);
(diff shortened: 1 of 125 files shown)


jrobinso:
Thanks but I'm not going to make this change, too ugly.

lbergelson:
It's fine if you don't want it, but it's a binary compatible change that should be completely transparent to downstream users of IGV as a library.  It also fixed a few bugs where getContig and getChr were defined to return different results.  It might be worth incorporating those.

lbergelson:
Actually, I'm not 100% certain that it's binary compatible, but it's definitely source compatible...  I'm not certain if replacing an overload with a default method is binary compatible...

jrobinso:
Thanks for the effort, I appreciate it,  I just don't want "getContig" everywhere.    There are 125 files changed here,  I'll see if I can find the bugs you allude to.

PR #596, opened 2018-10-29
Add setLogScale command to CommandExecutor, test for same.

hmenager:
the goal of this modification is to enable to switch between linear and log scales for tracks using IGV batch commands. The syntax is "setLogScale ['true' or 'false'] [trackName]". If trackName is left blank, range is applied to all.


jrobinso:
Looks  good.  Thanks for including a test.

hmenager:
thank you for merging it so quickly, that's awesome!

Issue #591, opened 2018-10-26
Expand insertions


jrobinso:
Expand insertions in alignment tracks by making a gap in the reference sequence.   Align insertions sequencing using cigar padding operators when present.

This has actually been implemented for over a year, but needs a little more work before releasing.

amwenger:
Two suggestions:
* Since the insertion markers are drawn globally on all BAM tracks when at least one BAM has an insertion, I think the "Show insertion markers" preference should exist only in the "Alignments" tab and not in the "Third Gen" tab.
* I have a hard time closing the expanded insertion.  Consider making the target zone for the "close" click bigger.  Perhaps allow clicking anywhere in the red bar that shows the range of the insertion rather than just in the purple triangle.

jrobinso:
OK, noted.  Its been a year since I worked on that,  but I do want to release it.   Please continue noting problems and suggestions.

kpalin:
This is a great feature (thanks) but there is a bug in situation where there are several alignment tracks (at least Third Gen tracks)  The insertions are extracted from all alignments but drawn only on one (I think the latest one added but not sure). Trying to expand such insertion throws nullPointerException. This with IGV 2.5.0 binary download.

```java.lang.NullPointerException: null                                                                                                                                                                                                                                                                                                 at org.broad.igv.sam.AlignmentTrack.renderExpandedInsertion(AlignmentTrack.java:663) ~[igv.jar:?]                                                                                                                                                                                                                            at org.broad.igv.ui.panel.DataPanelPainter.paintExpandedInsertion(DataPanelPainter.java:214) ~[igv.jar:?]                                                                                                                                                                                                                    at org.broad.igv.ui.panel.DataPanelPainter.paint(DataPanelPainter.java:89) ~[igv.jar:?]                                                                                                                                                                                                                                      at org.broad.igv.ui.panel.DataPanel.paintComponent(DataPanel.java:174) ~[igv.jar:?]                        ```

jrobinso:
Thanks for the report.   Does the error happen if you just load 1 track (the one the insertion is drawn for)?

kpalin:
Works. I now I can't even reproduce the exception. I'm of not much help now. I'll try to get a more reproducible bug report next time.

PR #606, opened 2018-11-20
Improve insertion markers / expanded insertions logic

amwenger:
Improve the insertion markers / expanded insertions logic:
* Connect the "Show insertion markers" preference to the render.
* Support "collapsed" mode BAMs in expanded insertion render.
* Fix an offset issue where the expanded insertion is drawn 3 px below its read.  I could not find which constants to tweak to the get the offset right so added a harcoded "-3".

Related to #591

### DIFF
--- a/src/main/java/org/broad/igv/prefs/Constants.java
+++ b/src/main/java/org/broad/igv/prefs/Constants.java
@@ -284,7 +284,8 @@ final public class Constants {
             SAM_QUICK_CONSENSUS_MODE,
             SAM_ALLELE_THRESHOLD,
             SAM_FLAG_LARGE_INDELS,
-            SAM_LARGE_INDELS_THRESHOLD
+            SAM_LARGE_INDELS_THRESHOLD,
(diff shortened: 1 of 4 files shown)


jrobinso:
Thanks.   I hope to finally release this around January next year.

jrobinso:
@amwenger I missed this in review, but this PR turned on insertion markers by default for everyone.  I'm not ready to do that yet.  Hopefully by the end of the month we will at last release the preferences window with platform specific settings and will do it then.   In the meantime I rolled that back, to see insertion markers in the released IGV you need to manually add this to the user.preferences file
```
SAM.SHOW_INSERTION_MARKERS=true
```

Just playing around with insertion markers they seem a bit buggy,  I need to find a couple of days to work with them.  For example clicking an open one doesn't always close it.   Also,  zooming out with an open marker sometimes does weird things.

amwenger:
Sorry, @jrobinso.  I did not note that in the PR summary, and I agree with having the markers off by default.

PR #688, opened 2019-09-13
Review and change README.md

brainstorm:
According to new updates after PR #620 got merged.

@davideby There is a gradle target that doesn't seem to be working referred to in README? Is it still used?:

```bash
$ ./gradlew createToolsDist

FAILURE: Build failed with an exception.

* What went wrong:
Task 'createToolsDist' not found in root project 'igv'.

BUILD FAILED in 2s
```

Can you take a quick peek into it please?


jrobinso:
@davideby Can you confirm that we don't use that gradle target anymore?  IIRC we don't build a tools distribution now.

davideby:
@jrobinso @brainstorm Yes, that's right.  We're not building igvtools at the moment.

PR #690, opened 2019-09-16
Cleaning up after dropping Java 8

davideby:
Drop separate Java 11 src folder.  No need since we're no longer
supporting Java 8 and now can have all src in the same place.
Removing old Java 8 launcher scripts.
Java 11 launcher scripts are now the main scripts.
Restore hidpi scripts.
Update README.md.
Remove some old JNLP resources.


jrobinso:
David, why are we modifying all the scripts, in particular removing "-Xmx1g"  etc.  Is that not necessary anymore?

scripts/igv.bat
```diff
@@ -1,4 +1,15 @@
+setlocal
 ::Get the current batch file's short path
 for %%x in (%0) do set BatchPath=%%~dpsx
 for %%x in (%BatchPath%) do set BatchPath=%%~dpsx
-java -Xmx4G -Dproduction=true -Djava.net.preferIPv4Stack=true -Dsun.java2d.noddraw=true -jar %BatchPath%\lib\igv.jar  %*
```
jrobinso, line 4: Where are we setting max memory now (e.g. -Xms4G)?

jrobinso: oh I see it

scripts/mac.app/Contents/Info.plist.template
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundleName</key>
-	<string>IGV_@VERSION@</string>
```
jrobinso, line 18: At one time @VERSION was replaced during the build to the current version, probably from the Jenkins project.  Do we not do that anymore?  That's normally where people go on a Mac to see the version.  They can still get it from the help menu, I think, so maybe this is not a big deal.

davideby: The Gradle build takes care of this now.  The value is supplied by Jenkins but Gradle does the substitution.  Look for ```filter (ReplaceTokens, ...``` in the build.gradle script.

jrobinso:
Thanks @davideby for thorough review and rewrite.

PR #691, opened 2019-09-18
Switch to Cloudtrail supported user access audit

brainstorm:
The enhanced flow does not support security audits (only the Role name appears on the logs). This change helps us a ton on things like accreditation and security audits (to see possible unauthorized user accesses to data).

This change requires just one additional line on `oauth-config.json`, pushed and documented on the public IGV-AWS backend blog post right now:

https://github.com/umccr-svc/site/commit/ef4efb5d0828f78516a89f9d0d0e7685eacff1fd

/cc @andrewpatto, change your `oauth-config.json` accordingly after this gets merged ;)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -138,9 +138,10 @@ dependencies {
         [group: 'org.netbeans.external', name: 'AbsoluteLayout', version: 'RELEASE110'],
 
         // Amazon deps
-        [group: 'software.amazon.awssdk', name: 'http-client-spi', version: '2.7.8'],
-        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.7.8'],
-        [group: 'software.amazon.awssdk', name: 's3', version: '2.7.8']
+        [group: 'software.amazon.awssdk', name: 'http-client-spi', version: '2.8.5'],
+        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.8.5'],
+        [group: 'software.amazon.awssdk', name: 'sts', version: '2.8.5'],
+        [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5']
     )
 
     testImplementation(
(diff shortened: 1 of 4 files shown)


mmatouk:
@brainstorm can you share some more details on the issue with enhanced flow? what is missing from cloudtrail logs. if you can send me example userIdentity field of your cloudtrail event that would be helpful.

brainstorm:
@mmatouk The main issue was that there was no way to track users using the enhanced flow. What you can see in CloudTrail logs looks like this (only the common Cognito Role ARN, nothing uniquely identifiable):

```
{type=AssumedRole, principalid=PPAL_ID:CognitoIdentityCredentials, arn=arn:aws:sts::ACCOUNT:assumed-role/Cognito_igvAuth_Role/CognitoIdentityCredentials, accountid=<ACCOUNT>, invokedby=null, accesskeyid=ASIABLABLABLA, username=null, sessioncontext={attributes={mfaauthenticated=false, creationdate=2019-07-03T06:16:49Z}, sessionissuer={type=Role, principalid=PPAL_ID, arn=arn:aws:iam::<ACCOUNT>:role/Cognito_igvAuth_Role, accountid=<ACCOUNT>, username=Cognito_igvAuth_Role}}}
```

Ideally, any of our **federated identity** users should appear there as "roman.valls@ourdomain.org" as users come from Google Apps via the `email` OpenID field. In a nutshell the **AssumedRoleWithWebIdentity** is not being forwarded in Cognito via the "Enhanced" auth flow.

From AWS support back then:

> (...) when switching to the old flow you will call AssumeRoleWithWebIdentity directly which will enable you to pass a ‘RoleSessionName’ that as far as I can tell will show up on the CloudTrail logs. I understand the frustration that comes from using our recommended flow but missing an important security feature, but I believe the workaround [switching to auth classic flow] will work for your use case and also won’t introduce too much overhead.

You can review AWS Support Case 6399785821 or ask away if you need more low level details. [I tried to update the official AWS Cognito docs on GitHub](https://github.com/awsdocs?&q=cognito) as I usually do when I find AWS bugs and unclear stuff, but it's not even there:

Skärmavbild 2020-02-28 kl  13 01 51

mmatouk:
Thanks @brainstorm - I'll review the case.

PR #692, opened 2019-09-18
Version bump htsjdk

brainstorm:
And kill the last (htsjdk) jar remaining under lib/. Tested and working fine :)


jrobinso:
@brainstorm I from time to time have to patch and release htsjdk from a custom branch in our repo to fix bugs while waiting for an htsjdk release.  I supposed if that situation arises again I can just undo this change temporarily.

brainstorm:
Yep, that's right, drop the jar under `lib/`, perhaps put the exclude back and that's it. Thanks for merging both PR so fast!

jrobinso:
@brainstorm o.k. hopefully that won't happen often.  It used to be fairly frequent.

Issue #705, opened 2019-10-03
S3 tree can be slow


jrobinso:
Opening this for discussion with @brainstorm.   We have a lot of buckets and a lot of files in our buckets.   The S3 picker tree can be slow to come up,  5 seconds or more.   This gets annoying if going to the menu more than once.   Is there anything we can do to speed this up, perhaps caching the tree so it is only slow the first time?  This would mean it doesn't get updated during a session.   This is not critical,  just a discussion point.

brainstorm:
Jim, IIRC the picker only retrieves the first level of the (dynamic) JTree (not all the files in the hierarchy)... but unfortunately we also noticed the slow sequential fetching and @reisingerf implemented parallel fetching of toplevel buckets... clearly that is not enough either :/

Caching (specially invalidation) is a hairy topic we didn't want to go too deep into so we left it as-is:

https://martinfowler.com/bliki/TwoHardThings.html

Would it be acceptable to pre-fetch that list when an AWS `oauth-config.json` is detected on app init and keep it in-memory? That way by the time the user goes to the picker it is already there and (ideally) not much caching is needed?... unless new buckets are created while the user session is active.

No strong opinions here, just brainstorming to find a simple solution that does not involve overengineering.

jrobinso:
Not sure how the pre-fetch would help because its when you go back multiple times that you notice it.  I agree at this stage a slow but robust implementation is better than a fast but complex one.

Maybe I can add a spinner while you're waiting, at least then it wouldn't appear frozen.

brainstorm:
Yes please spinner for now.

I meant to fetch it on IGV init time *and* cache it then all subsequent S3 file browsing from the same session would be fast... then invalidate when closing IGV. Again, not full solution since invalidation/refresh would have to be more involved anyway if users have really long sessions and S3 buckets are created meanwhile.

jrobinso:
@brainstorm Yes, that's exactly what I had in mind by "caching", sorry I wasn't more explicit.  Just cache it in memory for the duration of the session.   I think the initial delay when you first open the menu is o.k.,  its getting the delay repeatedly that gets annoying.  The "load from server" menu does that.

brainstorm:
This should be safe to close if all that was needed was to speeding up subsequent load dialog invocations.

The speeding up of the first invocation is a bit more involved since it has to happen in between:

* The login process comes back from the browser, right after the user has logged in.
* The user clicking in the Load from S3 submenu.

[...]

PR #711, opened 2019-10-11
Faster s3 load dialog

brainstorm:
Fixes issue #705


brainstorm:
@jrobinso Can you merge this PR before 2.7.0 release? It does fix subsequent S3 Load Dialog slowness (only the list buckets operation is slow, the rest is well handled by the dynamic JTree).

PR #716, opened 2019-10-20
Increase scroll increment in preferences dialog.

Lenbok:
The current scroll increment seems to be one pixel at a time, so scrolling the alignments tab in particular (which is significantly larger than the default window size) is ridiculously slow.

This bumps up the scroll increment so that it is usable.


jrobinso:
Thanks for the PR.  I generally grab the scroll handle and move it with the mouse, so I hadn't noticed this.

PR #722, opened 2019-11-03
Switch query- and subject-block would be better ?

CJ-Chen:
It seems that more users would use genome sequences as blast-subject instead of blast-query ?
For example, I might want to make a simple TBLASTN of one specific gene 
against the whole genome and visualize them with alignment data to check if there are any missing gene structure annotations.


jrobinso:
I don't see how this pull request is an improvement.  The terms used are common and taken from the blat documentation.

PR #723, opened 2019-11-04
Modify start/end position by dragging mouse

CJ-Chen:
It is a common case that users want to modify a bit of the ROI already defined.


jrobinso:
Yes this could be a good improvement, apologies for the delay I will try to get to it this week.  Thank you for the PR.

helgathorv:
Released in IGV 2.8.0

Issue #632, opened 2019-03-01
Suggestions for viewing long-read structural variant data


jrobinso:
From: https://groups.google.com/d/msgid/igv-help/3c79a4a8-f759-4380-a1d9-ff65083e99a4%40googlegroups.com?utm_medium=email&utm_source=footer

I've been using IGV lately to look at some Oxford Nanopore sequence data to find structural variants. IGV has been helpful, but there are definitely a few features (drawing on functionality for SVs from Illumina paired-end reads) which would be extra useful: 

My wishlist would be:

1. Setting to show hard-clipped bases as well as soft-clipped. 

Long-read aligners use a mixture of secondary and supplementary alignments (ie a mixture of hard and soft-clipping) to indicate split reads. These split reads are the ones that show structural variants, and the ones we want to see. Right now, IGV only allows soft-clipped bases to be shown, so that about half of the split reads don't show up their clipped portion when "show soft-clipped bases" is checked. It would be very helpful to have a setting to also show hard-clipped bases.

2. "View secondary/supplementary alignment region in split screen"

The "view mate region in split screen" functionality is invaluable for looking at SVs in Illumina reads. It would be equally invaluable to be able to do this for long split reads. One possible issue would be that long reads often have multiple additional alignments. Maybe in these cases it'd be best to open multiple views at once?

3. Extend "link supplementary alignments" to also link secondary alignments

As mentioned in 1), the choice between marking split alignments as secondary or supplementary seems to be somewhat arbitrary on the part of the aligner. At present, only supplementary alignments are being linked/coloured, which makes finding the secondary alignments tricky (since it involved hovering over reads one by one).

4. Create a "color alignments by secondary/supplementary alignment orientation"
The "color alignments by mate orientation" feature is also really useful; it would be great to have something similar for split long reads.

5. Highlight reads with clipped region greater than N
It would be very helpful to have some kind of visual cue (colour? read shape?) to find reads with large clipped sections. These sections often indicate SVs. Sometimes they are not aligned (eg if they're only a few hundred bp and in the middle of a repetitive region), but are clipped. Being able to quickly find and examine these reads would be really helpful.

oneillkza:
Thanks for opening this. Followed. Tag me if it would be helpful to make any clarifications!

helgathorv:
From @oneillkza at igv-help https://groups.google.com/d/msgid/igv-help/3c79a4a8-f759-4380-a1d9-ff65083e99a4%40googlegroups.com

[...]

PR #728, opened 2019-11-19
Add right-click menu items to BLAT soft-clipped sequence

amwenger:
As suggested in #632, add right-click menu items to BLAT soft-clipped sequence.

The menu items are only shown when the left/right clipping is longer than the "flag clipping threshold."  The BLAT track is labeled with the name of the searched read.

To test the functionality, you can load the hg19 BAM from https://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/AshkenazimTrio/HG002_NA24385_son/PacBio_SequelII_CCS_11kb/HG002.SequelII.pbmm2.hs37d5.whatshap.haplotag.RTG.10x.trio.bam and navigate to chr7:133,769,566-133,815,688.  

*Menu item*
image

*Example result*
image

### DIFF
--- a/src/main/java/org/broad/igv/sam/AlignmentTrack.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentTrack.java
@@ -1442,6 +1442,7 @@ public class AlignmentTrack extends AbstractTrack implements IGVEventObserver {
             addCopySequenceItem(e);
 
             addBlatItem(e);
+            addBlatClippingItems(e);
             addConsensusSequence(e);
 
             AlignmentBlock insertion = getInsertion(clickedAlignment, e.getMouseEvent().getX());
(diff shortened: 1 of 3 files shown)


jrobinso:
This looks good, I think the user can set the clipping threshold and I assume blat has some minimum string length.    For example I doubt you can blat "1".   We should probably have some absolute minimum.

jrobinso:
20 is the recommended minimum.  RE the Travis failure, its a Travis issue, have been seeing it a lot.  The tests pass.

jrobinso:
@amwenger I added an absolute minimum length of 20 bases, in general you get nothing back from UCSC blat for shorter sequences.   This works great, BTW, it will be in the next release.   I had an idea for the future,  suppose we inserted the blat alignment into the BAM track and connect it, with some different color or outline to distinguish it.  Something to think about.

amwenger:
Thank you, Jim.

> I added an absolute minimum length of 20 bases, in general you get nothing back from UCSC blat for shorter sequences.

Makes sense.  Perhaps take it one further step and always use 20 bp as the minimum, eliminating the connection to the `SAM_CLIPPING_THRESHOLD` parameter.  I suspect it will not be obvious to users that the option to BLAT clipping is controlled by that parameter, and I see no harm in offering "BLAT clipping" in the right-click menu as long as the clipping is large enough to make sense to BLAT.


>  I had an idea for the future, suppose we inserted the blat alignment into the BAM track and connect it, with some different color or outline to distinguish it. Something to think about.

I really like that idea.

helgathorv:
BLAT soft-clipped sequence - released in IGV 2.8.0

PR #739, opened 2019-12-09
added method to check for HTTP redirect before using URL

hervold:
This patch is meant to address a performance problem that occurs when IGV loads an indexed BAM file from a host that redirects to a different server.  In our particular case, this redirect operation is expensive, on the order of 3 seconds.  Given an indexed BAM file, IGV makes dozens of Range queries, each incurring a 3s penalty, which means the BAM file takes minutes to load where it should take seconds.

By patching IGV instead of htsjdk (which is the library actually making the Range requests), this should cover file types besides BAM + BAI.

Thanks for taking a look, and any feedback is appreciated!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -142,11 +142,12 @@ dependencies {
         [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5'],
 
         [group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0'],
-        [group: 'com.sparkjava', name: 'spark-core', version: '2.2']
     )
 
     testImplementation(
-        fileTree(dir: 'test/lib', include: '*.jar')
+        fileTree(dir: 'test/lib', include: '*.jar'), // first search on disk (old behavior), then maven repos
+        [group: 'com.sparkjava', name: 'spark-core', version: '2.2'],
+        [group: 'org.glassfish.jersey.core', name: 'jersey-common', version: '2.22.4']
     )
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -141,7 +141,8 @@ dependencies {
         [group: 'software.amazon.awssdk', name: 'sts', version: '2.8.5'],
         [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5'],
 
-        [group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0']
+        [group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0'],
+        [group: 'com.sparkjava', name: 'spark-core', version: '2.2']
     )
 
     testImplementation(
@@ -208,6 +209,7 @@ compileTestJava {
 
 test {
     inputs.property("moduleName", moduleName)
+
     doFirst {
         jvmArgs = [
             '-Dcom.sun.xml.bind.v2.bytecode.ClassTailor.noOptimize',
--- a/build.gradle
+++ b/build.gradle
@@ -139,7 +139,9 @@ dependencies {
         [group: 'software.amazon.awssdk', name: 'http-client-spi', version: '2.8.5'],
         [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.8.5'],
         [group: 'software.amazon.awssdk', name: 'sts', version: '2.8.5'],
-        [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5']
+        [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5'],
+
+        [group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0']
     )
 
     testImplementation(
--- a/src/main/java/org/broad/igv/util/HttpUtils.java
+++ b/src/main/java/org/broad/igv/util/HttpUtils.java
@@ -84,6 +84,8 @@ public class HttpUtils {
     // static provided to support unit testing
     private static boolean BYTE_RANGE_DISABLED = false;
     private Map<URL, Boolean> headURLCache = new HashMap<URL, Boolean>();
+    // remember HTTP redirects
+    private Map<URL, URL> redirectCache = new HashMap<URL, URL>();
 
     /**
      * @return the single instance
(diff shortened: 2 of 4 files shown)


jrobinso:
Thanks for the PR, this is a good idea, however it adds overhead for the vast majority of cases where no redirect is used.  The "head" request should be fast, but its still unnecessary for most cases.  Also, some servers don't support "head" and return an error.    Another strategy to explore would be to turn follow redirects of in HttpUtils, and catch this case as it occurs, caching the redirect url with the original url as a key.   Let me explore the feasibility of that and get back to you.  If that is successful the code here would move to HttpUtils.

BTW I would worry about this a little, getting stuck in an infinite loop due to server error.  In addition to the "!=null" test I think we need a counter and the equivalent of "too many redirects".

```
	    while( redirected != null ) {
		final_path = redirected;
		redirected = checkForRedirect(redirected);
	    }
```

You are brave to open ResourceLocator, which started small but has become a monster.

hervold:
Thanks for your quick response!  I can't argue with anything you said, and agree that catching the redirect and handling it sounds like a better approach.  I'll give that a shot, and if I can make it work, I'll update this PR.

Thanks again!

jrobinso:
OK,  I was going to look at that later today and give you some guidance.
 The place to start would be around line 833 of HttpUtils, where manual
redirects are already handled for special cases.     The other piece is
caching the URL for later use.   For this I would use the mapURL method
(line 127)  which already does various mappings, again for special cases.
  My idea would be to create a static cache,  and check it for redirects in
this method.

hervold:
You were right: this is a much simpler, cleaner approach than my first attempt.  However, I don't think I'm done yet, as the user might keep IGV running for a long time, and we don't want to cache a redirect for longer than the Cache-Control or Expires headers specify.

I'll add an expiration timestamp to the Map, based on the headers and defaulting to something reasonable, depending on whether we got a 301 (say 24 hours) or 302/307 (15 minutes).

jrobinso:
@hervold Yes, if there is information in the headers we should use it.   I had the same worry, that the redirect could change or become invalid.     If we have no information in the header maybe a short, as in 15 minutes, default expiration is a good idea.

hervold:
@jrobinso I've added some logic to parse redirect expiration headers, defaulting to 15 minutes.  "Cache-Control" takes precedent over "Expires," as specified in the docs:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

I'd like to add some tests, but I'm unfamiliar w/ Java test frameworks.  Can you point me to a good example?

Thanks!

jrobinso:
The tests are "JUnit",  mostly JUnit 4.  You could just add test methods to test/java/org/broad/igv/util/HttpUtilsTest.     A good simple example is URLUtilsTest.

We are currently wrapping up a release,  it will take a few days, after I will look at merging this.

hervold:
I added an instance of the Spark HTTP server for testing purposes, and added a simple test of the new redirect caching.  The basic test works: I can send a GET request to the local test server.  However, I ran into a problem and hoped you might have some advice.

I'm using the javax.ws.rs.core.CacheControl class to parse the Cache-Control header, as it's quite complex.  The code compiles just fine, but the test produces a ClassNotFoundException when attempting to instantiate the CacheControl object.  I suspect this has something to do w/ my test environment - gradle is missing the javax JAR file, perhaps?

jrobinso:
@hervold sorry for the long delay, we had a 2 week holiday shutdown here.  I will look into this, there are some things about Java 11 I still do not understand.  Perhaps @davideby can help here.   Give me a day or 2.

jrobinso:
A comment, I think spark-core should be a dependency of the test build (testImplementation), not "implementation".    We don't want to include that in the distributed bundles.   @davideby do you agree?

davideby:
Yes, that makes sense @jrobinso .  It will be at least a couple of days until I can take a look at this, though, so don't take this as definite.

There are some issues with jUnit-style testing under Java 11 that probably explain what @hervold is seeing.  No time to get into it just now but I'll do so when I can.

davideby:
Hi folks,

I finally got a chance to look at this.  @hervold, running with `./gradlew --info build` shows the full error + stacktrace:
`java.lang.RuntimeException: java.lang.ClassNotFoundException: org.glassfish.jersey.internal.RuntimeDelegateImpl`

A little web searching turns up [this solution](https://stackoverflow.com/questions/19452887/org-glassfish-jersey-internal-runtimedelegateimpl-not-found/20258438).  Basically Spark needs an implementation class and that thread suggests Jersey (there are probably other options; I don't know).  Also, @jrobinso is correct: these dependencies can be added under `testImplementation` to avoid shipping them in the distributions.

@hervold: Rather than have me do a PR on your fork, can you just change that in your gradle config as follows?
```
testImplementation(
    fileTree(dir: 'test/lib', include: '*.jar'), // first search on disk (old behavior), then maven repos
    [group: 'com.sparkjava', name: 'spark-core', version: '2.2'],
    [group: 'org.glassfish.jersey.core', name: 'jersey-common', version: '2.22.4']
)
```
Plus obviously remove spark from the compile `implementation` dependencies.

Everything passes after doing the above on my local repo.

hervold:
@davideby thanks for figuring that out!  The tests pass now.

davideby:
No problem @hervold!  I'll leave it to @jrobinso to review the PR from here...

jrobinso:
@hervold I've pulled this locally and will start reviewing tomorrow.

hervold:
@jrobinso - thanks for taking the time!  One intriguing possibility opened up by the spark HTTP server is eliminating the external dependencies in other tests, eg, instead of loading http://data.broadinstitute.org/igvdata/annotations/seq/hg19/chr1.txt in testGetContentLength, we could bring up a local HTTP server and use that.  If you'd like me to give that a shot, I'd be happy to.

jrobinso:
Thanks for the offer @hervold, it would be more involved than just moving files however.  Some of the tests use large files that we would not check into github.  I agree that ideally tests should not have external dependencies, but eliminating them would mean to rewrite the tests to use small files.    So its a big job, but if you want to contribute to that it would be welcome.

jrobinso:
@hervold I've reviewed this and it looks good, but I would like to make a change to avoid importing javax.ws  just for the purpose of parsing the Cache-Control header.   I've made this change in fact in this branch:  https://github.com/igvteam/igv/tree/hervold-igv.     I could do a PR to your branch, or alternatively just merge your PR then merge the branch above.

jrobinso:
@hervold and kudos for that unit test, its good stuff and as you point out can be used going for other tests involving loading by http.

hervold:
@jrobinso - you merged - that's great!  Thanks for being so responsive and helpful with this.  And you're right about adding javax as a dependency; now that you added CacheControl parsing, it's not as complicated as I'd feared.

PR #759, opened 2020-02-18
Fix AWS Glacier/Deep Archive loading of restored objects

brainstorm:
Up until now, error messages with AWS Glacier were limited to knowing the storage class.

When an object gets "thawed" (restored) from AWS Glacier, it holds the StorageClass (it does not transition to i.e standard). Instead, the header `x-amz-restore` should be parsed accordingly for the different lifecycle stages where the S3 object might be in.

While I was at it, I also added support for directly loading those S3 objects from `Load-> From URL` menu, to be consistent in UX/behaviour with the S3 Tree selector... this is a very **germane** change and goes along with this same PR!

Please @igvteam @jrobinso @davideby, would you please expedite this PR merge and release? Otherwise the Glacier object restoration will remain inoperative and confusing to users :-S

/cc @reisingerf @andrewpatto

### DIFF
--- a/src/main/java/org/broad/igv/ui/action/LoadFromURLMenuAction.java
+++ b/src/main/java/org/broad/igv/ui/action/LoadFromURLMenuAction.java
@@ -29,14 +29,10 @@
  */
 package org.broad.igv.ui.action;
 
-import htsjdk.samtools.util.Tuple;
-import org.apache.commons.lang3.tuple.Triple;
 import org.apache.log4j.Logger;
-import org.broad.igv.aws.S3Object;
 import org.broad.igv.exceptions.HttpResponseException;
 import org.broad.igv.feature.genome.GenomeManager;
 import org.broad.igv.google.GoogleUtils;
-import org.broad.igv.google.OAuthUtils;
 import org.broad.igv.prefs.Constants;
 import org.broad.igv.prefs.PreferencesManager;
 import org.broad.igv.ui.IGV;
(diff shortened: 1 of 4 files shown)


brainstorm:
@jrobinso Sorry for the last commit, no more are coming in unless you have something to add here.

This should be ready to merge, imho ;)

src/main/java/org/broad/igv/ui/action/LoadFromURLMenuAction.java
```diff
@@ -96,6 +98,20 @@ public void actionPerformed(ActionEvent e) {
...
+                        try {
+                            // If AWS support is active, check if objects are in accessible tiers via Load URL menu...
+                            if (AmazonUtils.isAwsS3Path(url)) {
+                                String bucket = AmazonUtils.getBucketFromS3URL(url);
+                                String key = AmazonUtils.getKeyFromS3URL(url);
+
+                                AmazonUtils.s3ObjectAccessResult res = isObjectAccessible(bucket, key);
+                                if (!res.getObjAvailable()) { MessageUtils.showErrorMessage(res.getErrorReason(), null); return; }
+                            }
+                        } catch (NullPointerException npe) {
```
jrobinso, line 110: Just a comment, do we know that the user hasn't logged in solely from the npe?  Could there be other reasons for an npe?

brainstorm:
Thanks for merging that fast, Jim! Looking forward for next release so I can tell our users to update.

I had exactly the same thought about that try-catch, but when I started to implement a `isAWSAuthenticated()` type of logic to make it better, it went down to a chicken-and-egg rabbit hole of auth states, so I don't want to complicate that bit, if you don't mind.

Promise I'll fix/refactor that part if it becomes a problem ;)

jrobinso:
I don't mind, just noting it, sometimes the simple solution is best.  Appreciate the PR.

jrobinso:
Oh on release,  I have a juicebox.js and igv.js release to get out first, then we can look at releasing this.  Maybe end of next week or week after, ping me if it drags on past that.

Issue #794, opened 2020-04-14
Google oAuth does not work properly if running multiple IGV instances


SHuang-Broad:
My IGV session needs access to 6 BAMs stored in Google Cloud buckets.

The nightly build (as of April 14th, 2020) asked for authentication code 6 times.

jrobinso:
I will look into it,  does the released IGV do this?  

Also, are you logging in from the "Google" menu?  This shouldn't matter, but see if that works as a workaround.

SHuang-Broad:
Nope. The release version runs smoothly.

> are you logging in from the "Google" menu?

I tried that, but it still asks me for a code and opens a web browser for generating that code.

I do have a release version IGV open at the same time, but I doubt that matters.

Thanks Jim!

helgathorv:
Actually, I have found that running multiple instances of IGV can confuse Google. Try having only one open.

SHuang-Broad:
Ah, I see. Maybe that's why. I'll git it a try.

SHuang-Broad:
With the snapshot/nightly build itself open, it no longer asks for authentication codes multiple times.

I guess this might be a google cloud problem then.

jrobinso:
So to confirm, the problem occurs with multiple IGVs open at one time, whether released or snapshot?   Is one of the IGVs just passively open, or do they both have Google cloud files open?  

Its not obvious why this should be.   Perhaps requesting an access token for the same app from the same ip address invalidates previous access tokens, but if one IGV is passive and not requesting tokens its not obvious how that would matter.    I'll see if I can reproduce it.

SHuang-Broad:
I opened the same IGV session with a latest release and latest snapshot, release version opened first, snapshot afterward. Then the snapshot version keeps asking for the code.

Can I open multiple release version instances (I mean, same version) at the same time?

jrobinso:
Well, apparently not when loading from Google buckets, based on your experience.     On a Mac you can't open multiple versions, what platform are you using?

SHuang-Broad:
Mac.
I was indeed trying to take a quick glance of a different site and don't want to loose the current site. That was why I wanted multiple instances of IGV open.

helgathorv:
You should be able to open the second IGV mac app by running the following from the command line:
open -n <app>
for example:
open -n /Users/helga/Applications/IGV_2.8.2.app 

This is true of Mac apps in general - it's not a specific IGV thing.

SHuang-Broad:
[...]

PR #798, opened 2020-04-21
Multiple IGV instances

brainstorm:
This pull request closes the loop on parametrising the IGV commands port via user preferences, as it is already via the `prefs.properties` settings:

```
$ cat ~/igv/prefs.properties
(...)
PORT_NUMBER=60151
PORT_ENABLED=true
```

This essentially fixes https://github.com/igvteam/igv/issues/794. As you know, we use Google as IDP, so this change should work as well for Google-only folks since all it's doing is parametrising the IGV command port(s). The usecase @umccr is to have two IGV instances connected to two separated AWS accounts:

One local IGV instance is configured on port `60151` by default, for prod.
The other is configured against port `60152`, for dev.

Hope this helps others.

/cc @ohofmann

### DIFF
--- a/src/main/java/org/broad/igv/google/OAuthProvider.java
+++ b/src/main/java/org/broad/igv/google/OAuthProvider.java
@@ -7,6 +7,7 @@ import org.apache.log4j.Logger;
 import org.broad.igv.Globals;
 import org.broad.igv.batch.CommandListener;
 import org.broad.igv.event.IGVEventBus;
+import org.broad.igv.prefs.PreferencesManager;
 import org.broad.igv.ui.util.MessageUtils;
 import org.broad.igv.util.AmazonUtils;
 import org.broad.igv.util.HttpUtils;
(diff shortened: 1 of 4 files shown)


brainstorm:
For more context, we keep the provisioning information for the IGV instances (both IDP/auth backend and IGV client) on Github. See the two `oauth-config.json.gz` and `prefs.properties` and accompanying `IGV` launcher scripts for an idea on the strategy):

https://github.com/umccr/infrastructure/tree/master/cdk/apps/igv/config

I think it's worth mentioning that the hardcoded `google_desktop` Broad's config:

```java
    private static final String PROPERTIES_URL = "https://s3.amazonaws.com/igv.org.app/desktop_google";
```

Should/could go away now if we make IGV more PROVISIONING_URL-centric from now on?

brainstorm:
After going through the code, I would actually advocate for migrating the whole preferences subsystem (existing prefs.properties) to (JSON) attributes under PROVISIONING_URL umbrella. 

I digress from the focus of this PR and that would be a significant change set, but I wanted to mention it as a possible IGV 3.x series consideration?

It would simplify config uniformity/compliance of IGV settings within an organisation, massive win IMHO.

jrobinso:
Looks good, but I don't understand the change to ParsingUtils.

brainstorm:
Oh, that was just a way to debug the GZipStream because I was having redirects when fetching the provisioning `.gz` file (and getting `Not a GZIP file` as a result, since it didn't follow redirects) :-S

I can remove it, it's just a comment anyway.

jrobinso:
Yea the comment is fine, but there is an import that wasn't there before.

jrobinso:
Thanks, clearly that is what should have been done to start with.

PR #802, opened 2020-05-03
Pull Request for Native File Dialogs for all environments. 

kelepiradam:
Hi 

These are my changes to IGV to use native file dialogs for all OS'es. Tested on macOS, linux and windows. 

Note: Some of the changes were made in my own repository for my own needs but they do not need to be merged. Sorry for any inconvenience it may cause. 

I also made some additions to ConsensusCalculator which allows user to set a different multiplier than 2 in the actual formula in case of using pooled samples and higher ploidy sequencing when calculating a consensus. This was a request from one of my colleagues. Default value will still be 2 but user can change it to any integer above 2. 

One last change is made to VariantTrack.java. Annotation softwares sometimes clog the whole variant info tooltip with tons of information where you want to see only some of them. I added a comma seperated list of VCF Info tags section where users can limit the items to be seen at Variant Info section which allows them to have a cleaner look at the annotation results. 

These last 2 changes may not be merged if not preferred.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,9 +1,9 @@
 # We override this file externally for production builds
 # Use special constants for non-production builds so that the jar doesn't change unnecessarily
 buildNumber=0
-version=2.8_Custom
-vendor=Sky
-timestamp=
+version=not_set
+vendor=not_set
+timestamp=not_set
 timestamp_fs=not_set
 dataServerUrl=http://www.broadinstitute.org/igvdata/$$_dataServerRegistry.txt
 development=false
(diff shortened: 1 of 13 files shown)


jrobinso:
Thanks @kelepiradam!  I will make every effort to review this by the end of the week.

jrobinso:
@kelepiradam I won't be able to accept this PR as a PR with the unrelated changes (changes not related to native file dialogs).    I can hand merge the file dialog bits tub it will be outside the flow of a PR, that's not how PRs work.

kelepiradam:
You are right and it was totally my mistake to not generate a seperate branch but to push all my changes to my own master fork. I will work on a seperate branch in my repo just to include native file dialogs and recreate a new pull request.

PR #821, opened 2020-06-19
Improve BAM rendering in alignments with indels

amwenger:
Improve BAM rendering logic to:
* Prevent overlapping blocks
* Suppress gaps that are less than 1 pixel at current resolution
* Consider "hide small indels" logic in outlining alignments

### DIFF
--- a/src/main/java/org/broad/igv/sam/AlignmentRenderer.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentRenderer.java
@@ -655,6 +655,8 @@ public class AlignmentRenderer {
         int arrowPxWidth = pixelLengthOnReference == 0 ? 0 : (int) Math.min(Math.min(5, h / 2), pixelLengthOnReference / 6);
 
         int prevBlockChromEnd = 0;
+        int outlineChromStart = blocks[0].getStart(),
+            outlineChromEnd = blocks[0].getStart();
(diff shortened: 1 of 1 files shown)


jrobinso:
@amwenger I'm concerned about un-intended side-effects here,  particularly "suppress gaps that are less than 1 pixel...", on viewers of short read alignments.   We typically want to show mismatches to reference, including 1-base indels,  at much coarser zooms than it would require to see them at 1 pixel resolution.    Why does "hide small indels" not accomplish this?

jrobinso:
I guess we want a 1 bp deletion visible at the same resolution as a base mismatch, by default.   We could have a different rule with experiment type = 3rd gen

amwenger:
Good point @jrobinso.  I will withdraw this PR and submit another, more focused one.  My main goal is to fix some apparent regressions in alignment rendering between IGV 2.7.2 and 2.8.6.

For example, for the small IGV session at https://dl.dnanex.us/F/D/zfG7yFjY0g1bG7GBQV9vkbYxYgZk9gGx3K6FJ0Gg/hg19_overlap.xml, IGV 2.7.2 renders more nicely than 2.8.6 using the same IGV session (hide indels <2bp):

**2.7.2**
igv-2_7_2

**2.8.6**
igv-2_8_6

I will submit a PR that addresses these three issues but does not hide indels based on pixel size.

PR #822, opened 2020-06-23
Fix alignment block rendering regressions related to "hide small indels"

amwenger:
Correct apparent regressions (from 2.7.2 to 2.8.6) in rendering of alignment blocks
with hidden indels. Correct:
1. empty space between alignment blocks at hidden deletions
2. shadows at block boundaries when drawing multiple times at same position
3. outlines at block boundaries for hidden indels

Attempt to improve closed PR #821. 

Session with example dataset: https://dl.dnanex.us/F/D/zfG7yFjY0g1bG7GBQV9vkbYxYgZk9gGx3K6FJ0Gg/hg19_overlap.xml

**2.7.2**
igv-2_7_2

**2.8.6**
igv-2_8_6

**This pull request**
pr


jrobinso:
Nice

jrobinso:
BTW I don't think this regression has ever been released.   There is a branch for 2.8.X releases and it seems fine there.  It only existed in the "master" branch.

Issue #833, opened 2020-07-31
"View mate region in split view" freezes IGV on distant reads


brainstorm:
Hello @jrobinso, @igvteam

I have been chasing an IGV freeze bug for a while now. **The regression appeared for sure on commit f9690c6047b92028503031f665dbbc0fe0aa25f4 on Jun 13th**. I bisected from release 2.8.3 (which works) an here's what I found, manually testing the sequence of steps reported by several staff and research members @UMCCR:

4fbb9df78c11234637bae4f2d7857685a5ecb5cb ... Feb 2?	Works
b15e0f0f5e82fbbf14cedef947eddb92a8745513 ...  Mar 7?	Works
4879b5e7eb3099c71ccd5dba004462a704ede677 ... Jun 1?	Works
ce921fd9e77c32b236589151ffcd02e7183ca562 ...  Jun 2?	Works
6979897cf66cf0a518a3f3e9d12661ac657e79c3 ... Jun 12?	Works
b0579ca76ddc8a2b7b79c52a14e5b4df2e52a8f5 ... Jun 12?	Works
**f9690c6047b92028503031f665dbbc0fe0aa25f4 ... Jun 13?	Hangs!**
dd8e3c0f5496990bb3626a3f0448063a0b5f9b72 ... Jun 13?	Hangs!
0f862063c4bcad58d639b1163c8e9322c0b1156e ... Jun 14?	Works but zoom on the right is not shown
b8855c53afb599bb9a33cbd0d1f102a491c0b1e1 ... Jun 20? 	Hangs!
...
All the way to `master` and prior releases **which all freeze/hang IGV completely**.

Would you mind revisiting that "async fix" change? I'll try to craft a minimal reproducer but it might take me a while, by reporting this in advance I hope you can figure it out faster what might be going on here.

/cc @ohofmann @reisingerf

jrobinso:
Thanks for the report, and for narrowing it down.   I'm able to reproduce this with a 1KG file, looking into it.

brainstorm:
Epic fast response, thanks @jrobinso, much appreciated! :)

jrobinso:
There was a thread lock contention.   I'm doing a lot of profiling now, and
looking at making some methods thread safe, to reduce the use of
synchronization generally.   This should give a performance boost in the
end.

However I think the fix I pushed returns us to the status quo.

brainstorm:
[...]

PR #855, opened 2020-10-19
Fix npe on AWS Storage Tier and treat the STANDARD tier status first

brainstorm:
Fixes observations made on https://github.com/igvteam/igv/pull/620#issuecomment-710450427 from two independent genomics labs.

This patch should also be cherry picked for next releases as well and the error unfortunately [affects IGV versions ranging from 2.7.1 to 2.8.0 as pointed out by @welcomege](https://github.com/igvteam/igv/pull/620#issuecomment-710696580).

Alternatively, I can also open a separate pull request against `master`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,10 +131,10 @@ dependencies {
         [group: 'org.netbeans.external', name: 'AbsoluteLayout', version: 'RELEASE110'],
 
         // Amazon deps
-        [group: 'software.amazon.awssdk', name: 'http-client-spi', version: '2.8.5'],
-        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.8.5'],
-        [group: 'software.amazon.awssdk', name: 'sts', version: '2.8.5'],
-        [group: 'software.amazon.awssdk', name: 's3', version: '2.8.5'],
+        [group: 'software.amazon.awssdk', name: 'http-client-spi', version: '2.15.9'],
+        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.15.9'],
+        [group: 'software.amazon.awssdk', name: 'sts', version: '2.15.9'],
+        [group: 'software.amazon.awssdk', name: 's3', version: '2.15.9'],
     )
 
     testImplementation(
--- a/src/main/java/org/broad/igv/util/AmazonUtils.java
+++ b/src/main/java/org/broad/igv/util/AmazonUtils.java
@@ -246,8 +246,15 @@ public class AmazonUtils {
         String s3ObjectStorageStatus = null;
         String s3ObjectStorageClass;
 
+        // Simple "null" case. The object is directly accessible in
+        // STANDARD, INFREQUENT_ACCESS, INTELLIGENT_TIERING
+        // or any other "immediately available" tier.
         s3Meta = AmazonUtils.getObjectMetadata(bucket, key);
-        s3ObjectStorageClass = s3Meta.storageClass().toString();
+        if (s3Meta.storageClass() == null) {
+            res.setErrorReason("Object is in an accessible tier, no errors are expected");
+            res.setObjAvailable(true);
+            return res; // nothing else to check, return early
+        }
 
         // Determine in which state this object really is:
         // 1. Archived.
@@ -260,6 +267,8 @@ public class AmazonUtils {
         //
         // Possible error reason messages for the users are:
 
+        s3ObjectStorageClass = s3Meta.storageClass().toString();
+
         String archived = "Amazon S3 object is in " + s3ObjectStorageClass + " storage tier, not accessible at this moment. " +
                 "Please contact your local system administrator about object: s3://" + bucket + "/" + key;
         String restoreInProgress = "Amazon S3 object is in " + s3ObjectStorageClass + " and being restored right now, please be patient, this can take up to 48h. " +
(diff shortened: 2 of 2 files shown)


jrobinso:
Looks good.

brainstorm:
Fantastic, thanks Jim! Will we see a release `2.8.11` following up? I'll update the umccr blogpost so that users don't fall into this trap anymore, thanks for the patience!

brainstorm:
Also, do you want me to open another PR to apply this patch to `master`? Let's make sure it doesn't get lost on 2.8.x branch and resurfaces again later :-S

jrobinso:
Hi @brainstorm yes well will do a 2.8.11 release for this.  Possibly this week but more likely next.   I have merged this to master, which involved resolving some conflicts,   I think it is o.k. but if can review or test it that would be good.  The commit  id is 39658c6e  (last commit).

brainstorm:
Thanks Jim! Did you really push that commit to `master`? I can only see "fix unit tests" one from 5 days ago right now :-S

jrobinso:
Oh you're right, committed not pushed, its pushed now.

brainstorm:
Jim, just tested it on `master` too, LGTM, ready to ship a 2.8.11 release I reckon ;)

jrobinso:
@brainstorm What platforms are you testing on?  Something in the 2.8.11 release is causing IGV to hang in Windows 10.  Not an issue on MacOS.  Its not necessarily this PR, but there is very little else that differs.    The coding changes here look innocuous, but there are a few AWS library updates.     Anyway my question is whether or not you have tested this on Windows 10 before I jump into myself.

brainstorm:
I mostly test it on OSX since it's what most of the staff uses... I'll install it on windows and give it a go soon today to see what's going on, thanks for the heads up and your testing!

brainstorm:
@jrobinso Sorry, but I cannot reproduce :/

I just tested it on Windows 2019 (should be similar to Windows 10 I reckon?) and it seems to work just fine?

Skärmavbild 2020-11-02 kl  21 46 58

jrobinso:
I did not know until just now that there was a Windows 2019, but it
definitely doesn't work on Windows 10.    It does work on Windows 7.
However it's not necessarily the PR.   I have a Windows 10 machine
available and I plan to set up a dev environment to dig into this today.

jrobinso:
@brainstorm,  I should clarify "not work".   It is noticeably slower, and can appear to hang when loading bam files on Windows 10.   There are no coding changes that could cause this, so I'm looking at the Maven updates.   Do you have any idea what this library does?   http-client-spi

brainstorm:
I just tested without that dependency and all seems to work fine, so it should be safe to remove... is it the culprit? Otherwise I'm thinking of sync/async behaviors with http clients now that there's better threading (changes you were working a short while ago)? I.e:

https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/client-configuration-http.html

It's curious that just windows 10 is exhibiting this... different OS threading handling I (wild) guess?

jrobinso:
Yes its very curious, the coding changes between 2.8.10 and 2.8.11 are
minimal.   However there are a substantial number ro changed jars in the
final build, probably pulled in as AWS jar dependencies.

Your intuition about threading could well be the case, but what are you
referring with  re  "... now that there's better threading ..."?
Nothing's change in our code between 2.8.10.   I don't understand the
relevance of the link you referenced,  we don't use ApacheHTTPClient.

Since the dependency referenced apparently isn't needed are there others
that could be removed?

brainstorm:
> but what are you referring with  re  "... now that there's better threading ..."?

I meant some sync/threading changes that went through on https://github.com/igvteam/igv/issues/833#issuecomment-667446708 ... most probably unrelated, but just trying to rewind commits on my memory that can be the source of the current issue.

> I don't understand the relevance of the link you referenced,  we don't use ApacheHTTPClient and I don't recall working on this issue.

IGV might not, but somewhere inside AWS Java v2 SDK it could, that's why I was asking more details about at which point do you see the slowdown... right after trying to load a file (could be S3 pre-signing), downloading the reads (could be threads issue), reading from a local BAM file (could be htsjdk BAM reader)? etc...

> Since the dependency referenced apparently isn't needed are there others that could be removed?

Those three remaining AWS deps on `build.gradle` should be absolutely essential: STS, Cognito and S3 it's all this IGV-AWS extension does really.

jrobinso:
I need to break for a few hours.    This should have nothing to do with S3 pre-signing,  I'm reading from public buckets.  AWS isn't even configured on my setup.    Now I have a new challenge, I could reliably hang it earlier today now not at all.   It hangs, when it hangs, as alignments are read.   I say "hang", but the user who first reported this said it did load eventually, just very slowly.    I don't think this has anything to do with the PR or new libraries, except perhaps exposing a bug waiting to happen.    Anyway that's all for now, might have an update later this evening (or whatever it is for you).

BTW one curious phenom,  when it hangs the status message on the lower left of the status bar changes to "Iterating...".   What's curious about that is  I can't find anywhere in the code that sends such a message.   So first thing in the next go at this will be to determine where that's coming from.

jrobinso:
I'm not really sure what the 2.8.11 problem was, but it wasn't this PR.  Now released in 2.8.12.

brainstorm:
Thanks Jim! Intermittent bugs are the most annoying, we'll keep an eye on our Windows 10 users and see if they hit that issue.

I'll relay that release announcement to my peers, much appreciated ;)

Issue #856, opened 2020-10-22
Support reading BigWig and other file formats from S3 time-gated presigned urls 


mdug-gfb:
We've set up the Amazon IGV backend and it seems to be working well for most files, BED files or BAMs open without issue, however when we try to open bigwig files, we get an error of "file not found" or similar. This is true on both Windows and macOS, running IGV 2.8.10 

Looking in Fiddler while running trying to open the BED files, I can see traffic going to Amazon over port 443 and responses coming back, and the BED file is loaded in IGV. When trying to open a BW file I can see the request being sent for the file and a 200 response from Amazon but we get the "file not found error" in IGV. 

The igv log shows a java.io.FileNotFoundException. I can post the full log if that is helpful.

awilcz:
We're having the same issue. Local IGV, loading files from S3 bucket. BAM/BAI works fine, bigwig doesn't with "no such file or directory" error

jrobinso:
@awilcz @MarkSDuggan  Could you both elaborate?  All of our publicly served data is loaded from S3 buckets, including many bigwig files.  For example,  this, which you can load using "File > Load from URL"  https://s3.amazonaws.com/igv.org.test/data/ENCODE/ENCFF000ARJ.bigWig 

I haven't had time to test this yet but I assume you are using the backend setup described at the UMCCR site, is that correct?   @brainstorm Are you serving any bigwig files there?

brainstorm:
@jrobinso No, we are not using bigwigs via AWS-IGV @umccr... and I recall the reason behind not covering all file formats with the IGV-AWS support, see the specific discussion under https://github.com/igvteam/igv/pull/620#discussion_r317592218.

Namely:

> The fixes you have will work for alignments and many feature formats, but will not protect all file formats (...) It should probably be note that there could be a problem with formats other than bam and tribble type features. Specifically fasta, bigwig/bigbed, and tdf. Of course all formats will work until the url expires, as will as any non-indexed type format for which the entire file is read at once.

Correct me if I'm wrong but essentially we chose not to implement more than BAM/VCF because (currently) every format requires refreshing the reader instead of having a common "abstract" reader that can be used for all, IIRC.

/cc @reisingerf @ohofmann

jrobinso:
Thank you @brainstorm ,  I recall that now.

So I think for now the answer is IGV does not support all formats.   We should probably document what is supported.    Would that be BAM and VCF?   Anything else?  

[...]

PR #865, opened 2020-11-11
Add functionality to get AWS credentials from the Default Credential Provider

bbreton:
This adds the ability to get AWS credentials for accessing s3 from any of the supported methods on the AWS credentials chain.

https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default

This makes it so Cognito no longer needs to be setup should the organization using AWS not wish to and/or for minimal uses Cognito is overkill.  It also will allow for the instance profile credentials (e.g. you start up an EC2 instance and that instance has the credentials baked into it) to be used.  Overall it increases flexibility for uses that are not organizations that need cognito and/or want to run IGV with s3 access on AWS EC2 instances.

This implementation is a bit of a hack using the OAUTH code/path even though it isn't OAUTH.

My Open Questions:

Are you interested in adding this functionality?

What changes would you like to see to merge this and/or is the current method acceptable?

How long should I sign the S3 urls for? Right now it is set for the max (7 days).  This can be set to the default of 15 minutes or any other time including adding a config parameter for the user to set it.

What text should I put for the settings checkbox and is the current location good? I don't love my current text there, but trying to convey exactly what it is without getting long is difficult.

Thank You

### DIFF
--- a/src/main/java/org/broad/igv/google/OAuthProvider.java
+++ b/src/main/java/org/broad/igv/google/OAuthProvider.java
@@ -12,6 +12,7 @@ import org.broad.igv.ui.util.MessageUtils;
 import org.broad.igv.util.AmazonUtils;
 import org.broad.igv.util.HttpUtils;
 import org.broad.igv.util.JWTParser;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
 import software.amazon.awssdk.services.sts.model.Credentials;
 
 import java.awt.*;
@@ -225,11 +226,13 @@ public class OAuthProvider {
 
             if (authProvider.equals("Amazon")) {
                 // Get AWS credentials after getting relevant tokens
-                Credentials aws_credentials;
-                aws_credentials = AmazonUtils.GetCognitoAWSCredentials();
+                if (!PreferencesManager.getPreferences().getUseAwsProfile()) {
+                    Credentials aws_credentials;
+                    aws_credentials = AmazonUtils.GetCognitoAWSCredentials();
 
-                // Update S3 client with newly acquired token
-                AmazonUtils.updateS3Client(aws_credentials);
+                    // Update S3 client with newly acquired token
+                    AmazonUtils.updateS3Client(aws_credentials);
+                }
             }
 
 
--- a/src/main/java/org/broad/igv/prefs/Constants.java
+++ b/src/main/java/org/broad/igv/prefs/Constants.java
@@ -257,6 +257,7 @@ final public class Constants {
 
     // OAuth provisioning
     public static final String PROVISIONING_URL = "PROVISIONING.URL";
+    public static final String USE_AWS_DEFAULT_CREDENTIALS = "PROVISIONING.USE_AWS_DEFAULT_CREDENTIALS";
 
     // Experimental
     public static final String SCORE_VARIANTS = "SCORE_VARIANTS";
--- a/src/main/java/org/broad/igv/prefs/IGVPreferences.java
+++ b/src/main/java/org/broad/igv/prefs/IGVPreferences.java
@@ -375,6 +375,10 @@ public class IGVPreferences {
         return get(PROVISIONING_URL);
     }
 
+    public Boolean getUseAwsProfile() {
+        return getAsBoolean(USE_AWS_DEFAULT_CREDENTIALS);
+    }
+
     public String getPortNumber() {
         return get(PORT_NUMBER);
     }
--- a/src/main/java/org/broad/igv/util/AmazonUtils.java
+++ b/src/main/java/org/broad/igv/util/AmazonUtils.java
@@ -8,10 +8,13 @@ import org.broad.igv.Globals;
 import org.broad.igv.aws.IGVS3Object;
 import org.broad.igv.google.OAuthProvider;
 import org.broad.igv.google.OAuthUtils;
+import org.broad.igv.prefs.PreferencesManager;
 import org.broad.igv.ui.IGVMenuBar;
 import software.amazon.awssdk.auth.credentials.AnonymousCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider;
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.core.exception.SdkServiceException;
 import software.amazon.awssdk.regions.Region;
@@ -36,6 +39,7 @@ import java.net.URI;
 import java.net.URL;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
@@ -84,6 +88,16 @@ public class AmazonUtils {
             AWSREGION = Region.of(GetCognitoConfig().get("aws_region").getAsString());
         }
         return AWSREGION;
+    }
+
+    /**
+     * Returns the AWS credentials
+     *
+     * @return returns the credentials based on the AWS STS access token returned from the AWS Cognito user pool.
+     */
+    public static AwsCredentials GetProfileAWSCredentials(){
+        ProfileCredentialsProvider profileCredentialsProv = ProfileCredentialsProvider.create();
+        return profileCredentialsProv.resolveCredentials();
     }
 
     /**
@@ -167,7 +181,8 @@ public class AmazonUtils {
 
 
     /**
-     * Makes sure the S3 client is available for bucket operations and/or generation of pre-signed urls
+     * Makes sure the S3 client is available for bucket operations and/or generation of pre-signed urls using
+     * Cognito Credentials
      *
      * @param credentials AWS credentials
      */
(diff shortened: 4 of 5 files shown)


jrobinso:
@bbreton Thanks for the PR.   Yes I am interested, but it will be a few weeks before I can turn my attention to this.   I also have some outstanding S3 issues to contend with, such as #856,  which may or may not impact this PR.   

I would be interested to get input from @brainstorm

brainstorm:
Thanks @jrobinso for the heads up, appreciate that.

I **feared** that this type of PR would come along sooner or later since I got this type of feedback from some users and here it is... so I think it's worth clarifying and re-iterating the need for proper **security and good practices in IGV desktop when working in conjunction with AWS services**.

I get what @bbreton is trying to do here and it **sounds good in principle, but it's bad (dangerous) in practice**, let me explain why:

1. Cognito is a [**free** (for less than 50.000 users)](https://aws.amazon.com/cognito/pricing/) AWS service that provides temporary, time-gated (STS) credentials for non-IAM users. One might think it's overkill but [its use is fairly common practice and provides pretty good security guarantees.](https://transcend.io/blog/restrict-access-to-internal-websites-with-beyondcorp)  **users are known to heavily misuse and underestimate the security risks behind IAM credentials**. Which brings me to the next point.
2. IAM users should not be used like that. [Please carefully read the Stage 5 of AWS Security Maturity Roadmap](https://summitroute.com/downloads/archive/aws_security_maturity_roadmap-Summit_Route_2019.pdf) by Scott Piper (@0xdabbad00).
3. Statements like *"How long should I sign the S3 urls for? Right now it is set for the max (7 days)"* defeat the purpose of an adequate **time-gated**, secure, access to sensitive data, **widening the potential damage if a data breach occurs**. Also, setting 7 days as the maximum ignores the fact that only IAM users can provide that. Other mechanisms like "assume role"-like creds are limited to 12h or less, so that hardcoded max limit is already misleading for other secure use cases.
4. *"Overall it increases flexibility for uses that are not organizations that need cognito and/or want to run IGV with s3 access on AWS EC2 instances."* ... it does increase flexibility, but **who's willing to support that flexibility in terms of user support while reasonably guaranteeing its security on most third party deployments**? Certainly it will not be me if this gets merged as it is now.

Please don't get me wrong, I understand that supporting Instance Profiles and the Credential Provider Chain is a very convenient way to have **IGV desktop running on an EC2 instance, avoiding the need to install clients on user endpoints (a pain I'm well aware of myself)**... OTOH, it would defeat the purpose of having a nimble **pay-for-what-you-use S3-only setup**, as it is now.

Zooming out from my security concerns and getting some perspective on this topic, @bbreton's proposal and potential use cases involve running EC2 instances (if I understood correctly). That would require users to connect via X11/RDP/[VDI](https://aws.amazon.com/workspaces/)... **At this point I'd really step back and think: isn't this approach overkill when one just wants to use IGV DESKTOP**... specially when there's [igv.js](https://github.com/igvteam/igv.js) which is 0-heavy server side?

Full disclosure, I'm not claiming our (@umccr) implementation is perfect by any stretch of the imagination. For instance, [I'm still uneasy with the current JWT validation scheme implemented in IGV's desktop codebase, which I raised before on the original PR (which would need a proper and professional security audit)](https://github.com/igvteam/igv/pull/620#issuecomment-521872443). I don't want to come across as the typical IT security alarmist, but **please be aware that IGV is frequently used to access clinical genomics (patient) data so it should deserve higher scrutiny than most (rushed, security-as-an-afterthought) scientific software**, IMO.

Please, please, @bbreton, don't take this as personal criticism/grilling. I'm genuinely flattered that some code we wrote "out there on the internet" gets used by great people, so don't be discouraged by this analysis.

bbreton:
@brainstorm 

First off would you prefer I pull this request while the security parts are considered? I can't really put the genie back in the bottle, but deleting this probably avoid it hitting google and becoming too widespread. I added my email to github so if you want to contact me directly feel free.

The use case I specifically have is:

I have scripted IGV to take a large amount of IGV snapshots that I distribute out using aws batch to snapshot 1000+ locations at once. These snapshots can allow the rapid review of many locations such as is needed when trying to build a true positive database for whatever you are trying to find.

If anyone else has any ideas that would let a user review a diverse set of genomic locations on bams stored on s3 with less than a second load time per location I would be open to it. My current approach is simple, but for user time spent waiting for things to load I haven't found a better way.

I ended up doing a docker container that runs IGV combined with proxying my s3 bucket to do this on AWS batch, but that was a lot of effort and not very nice overall which is why I was looking for a way to give the EC2 instance/container permissions.  

One option would be to limit this feature to InstanceProfileCredentialsProvider and ContainerCredentialsProvider.


I agree with your concerns about the signed url.  My setting it to 7 days initially was somewhat for my environment/situation where it would be difficult to impossible for a signed url to leak.  I wasn't certain what the best trade-off for functionality vs. security would be which is why I pointed it out as a question I had.

`IAM users should not be used like that. Please carefully read the Stage 5 of AWS Security Maturity Roadmap by Scott Piper (@0xdabbad00).`
I believe that they would consider AWS SSO with assume role and using it to generate a temporary id/key as compatible with that roadmap.  So yeah this pull request enables insecure things, but it also opens up other options that for a given organization may be easier to secure depending on existing infrastructure.


For the more general concerns about security some of this gets into more security philosophy which I suspect my thoughts are somewhat driven by my personal experiences on small teams/companies with minimal IT organizational support and all the users being relatively sophisticated.  I am fully aware that IGV is frequently used for clinical genomics data and have dealt with PHI compliance including HIPAA.

My thoughts are:

The AWS credential chain can be used securely with proper setup.  The exact combination that best suits an org's security, existing infrastructure, regulatory and other requirements will vary.  That said an org can always do things like proxy s3 like I did which would allow them to use whatever auth they wanted so IGV supporting other methods isn't absolutely necessary.

I agree Cognito is probably the best way that a user can't easily drastically mess up security, but as above it isn't the only method to secure access.

If someone who doesn't understand why using the basic AWS IAM credentials at the very least need to be carefully secured and given minimal permissions and ideally transition to something with SSO or cognito especially if dealing with PHI has the ability to generate/give them out I feel that environment is already insecure and I'm not sure IGV not supporting one way of being insecure is going to help.

jrobinso:
@bbreton This is off topic,  nothing to do with security, but have you checked out igv-reports  (github.com/igvteam/igv-reports)?    It creates standalone html pages for variant review with data from the bam, vcf, and whatever else embedded right in the page and viewed with igv.js.   It is the ultimate in security because nothing is on the network, its literally a single html file.   I think Brian Haas uses this with 1000s of sites,  IIRC 10s of thousands,  I don't really know what the practical upper limit is but there are workarounds when we meet it.   My vision for that was to replace the static screenshot use case, which is a common one.   I haven't had time to promote it really,  but please check it out.    I would be happy to setup a video meeting to demonstrate it.

RE security and IGV in general,  users and organizations do all sorts of things with IGV and many other programs, we aren't  the secops police.  My goal would be to give organizations the tools to make IGV secure but its up to them to implement what is secure for their organization.    The most common scheme for accessing S3 resources with IGV, by far, is using signed URLs generated external to IGV.   Many organization with their own secops team feel this is secure. We on the IGV team are not requiring incognito, recommending icognito, nor any other particular scheme, we simply support it through UMCCR's contribution.  

@brainstorm s critique is more about best practices, as he sees them, that's not a debate I have expertise on or will engage in.   If there is a security hole in this PR (I haven't even looked at it yet) that would need to be addressed, but if it merely enables an insecure practice if misused that is another matter.   We also support basic auth,  which in general is a very bad authentication scheme but good enough for some organizations and usages.   For example locally here labs have used basic auth just to keep their data hidden from other labs until they publish.  

I don't know where this PR falls and won't have time to study it until probably post Thanksgiving, but no you don't need to pull it.

bbreton:
> @bbreton This is off topic, nothing to do with security, but have you checked out igv-reports (github.com/igvteam/igv-reports)? It creates standalone html pages for variant review with data from the bam, vcf, and whatever else embedded right in the page and viewed with igv.js. It is the ultimate in security because nothing is on the network, its literally a single html file. I think Brian Haas uses this with 1000s of sites, IIRC 10s of thousands, I don't really know what the practical upper limit is but there are workarounds when we meet it. My vision for that was to replace the static screenshot use case, which is a common one. I haven't had time to promote it really, but please check it out. I would be happy to setup a video meeting to demonstrate it.

Thanks.  I will take a look at it.  It certainly sounds like it will meet the requirements and would be easier than the scripting/snapshotting.  

That said I have gotten some feedback (including my own impressions) that many prefer the desktop app view, but much of that is just what users are used to and also may not apply to bulk review vs. careful examination of a few sites which are very different use cases.

jrobinso:
@bbreton To be clear I was not suggesting igv-web / igv.js could replace the desktop app, at least not now, I was suggesting it as a possible replacement for browsing static screenshots.

bbreton:
My current future plans for this pull are:

1. Add the capability to use AWS assume-role along with the ability to enter/use a MFA token if the assume-role requires it.
2. Add the capability to specify a credentials profile when using the aws credentials.
2. Figure out a better way to do the settings given I'm up to 4 settings needed.  Perhaps a checkbox similar to the google settings with a separate settings config dialog makes sense here.
3. Give my code a more overall refactor to make it cleaner and a bit more separated from the cognito/OAUTH code to avoid any risk that a bug in this new code could compromise the cognito code's security.

brainstorm:
For snapshotting, I'd either use https://github.com/brentp/jigv or https://github.com/igvteam/igv-reports, TBH... the former seems to be a bit more in line with what you were looking for, @bbreton.

jrobinso:
For static snaphsots the former, if you want to interact with the data the latter.

markotitel:
does it make sense if IGV has `access key` / `secret access key` fields?
User can get the keys from AWS admin and put it in the app.
The keys can have read only access to the bucket.
And voilla.
Java sdk should have `getObject()` on S3.

brainstorm:
@markotitel Please use the already officially supported AWS S3 via Cognito as mentioned in the README. Happy to assist as I did recently for a third party user: https://www.biostars.org/p/9506536

I understand that this PR seems "simple" but from a systems/security perspective it can become a nightmare fast, I'm glad it's finally closed and I hope it doesn't resurface as I noted in #1084.

jrobinso:
To be clear to everyone the IGV team is not taking a position on the discussion above between @bbreton and @brainstorm , the issue here is closed due to staleness.

jrobinso:
@bbreton At long last I have a solution for this I think, see PR (from myself) #1139.    Its a very simple implementation,   IGV will support the AWS SDK default credential chain.    Actually that's overstating it,  IGV itself knows nothing about credentials, it will simply create signed URLs with default values,  if the SDK finds credentials it will use them.   IGV itself does not copy or store any credentials.    Well now I've made a simple thing complicated.

In summary this means, for example, if access keys are stored where the AWS tools expect to find them they will be used, for details see the links in this PR.    Apologies that it took me so long to get to this.

Issue #836, opened 2020-08-05
[Feature request] Ability to set/control the sequence track strand in Batch Script/Socket


bbreton:
https://software.broadinstitute.org/software/igv/sequence_track_options

In the sequence track I would like to be able to set the strand (+/-) which is currently controlled in the GUI only with "Flip strand" when using the Batch Script/Socket control methods.

My specific use case for this issue is I want it to set it to the correct strand when using a Batch Script/Socket control doing snapshots so it has the right direction. Explicitly setting the strand makes more sense for this use case, but I could work with the ability to flip it as long as I could tell the starting state in case that is easier to do or fits other uses better.

Also (perhaps this should be a separate issue) the current strand of the sequence track isn't saved in the Save Session xml file as I looked into using loading a session as a workaround for no Batch Script command to control it. I feel like it would make sense for it to be saved in the session.

My testing was done with 2.8.9.

Thank you IGV team for all your work and contributions to the community!

jrobinso:
Yes a command would make sense.   The second issue is a bug, that should be in the session.   Will look at both.

jrobinso:
@bbreton is there anything still outstanding (i.e. left to do) here?

PR #868, opened 2020-11-18
Git Issue #836 SequenceTrack commands, preferences and session work

bbreton:
I'm still working on a commit for the AWS work to refine it a bit, but in the meantime I had a near term use for getting this working with the commands (assuming I don't end up switching to igv.js as discussed elsewhere). Either way I think this will be useful for the general community even if I don't end up using it.

Open questions:
setShouldShowTranslationCommand could probably be cleaner... I had a hard time getting it to properly refresh and ended up with what worked.

Should I add some tests for this?  I looked and the preferences, sessions and commands don't seem to be tested all that much currently.

Is there a way for me to update the html doco at https://software.broadinstitute.org/software/igv/PortCommands once this is released?

Features:

Adds saving the sequence track last strand to the preferences
Adds saving the attributes SequenceTrack.shouldShowTranslation and sequenceTranslationStrandValue to the igv_session.xml file for the Reference sequence xml.
Adds a command for batch scripting and socket control to set the strand as well as shouldShowTranslation.

Details:

New commands:

setSequenceStrand: valid values are case insensitive: +, POSITIVE, -, NEGATIVE
setSequenceShowTranslation: valid values are case insensitive: TRUE, FALSE

New preferences (under hidden since it is controlled with the clickable arrow):

SEQUENCE_TRANSLATION_STRAND: valid values are case insensitive: +, POSITIVE, -, NEGATIVE

New igv_session values:

Panel.Track.org.broad.igv.track.SequenceTrack.shouldShowTranslation: set show/hide translation
Panel.Track.org.broad.igv.track.SequenceTrack.sequenceTranslationStrandValue: set translation strand


jrobinso:
As it happens I'm about to start a round of IGV work.    I'll add this PR to the associated "Milestone" and get to it asap.

WRT port commands documentation,  we are going to move that and all doc from the internal site to a hosted markdown site somewhere, where we can accept PRs,  for now I'll have to add doc manually.

jrobinso:
Overall looks fine, here are a few comments
* This command and method is specific to SequenceTrack,  I'd prefer to not add these methods to the Track interface & consequent do-nothing implementations,  casting to SequenceTrack is o.k. here, or rather the lesser of 2 evils.
* Let's not rely on the sequence track name,  an "instanceOf" test would be more robust.  ```t.getName().equals("Reference sequence")```
* The batch/port command names are a little verbose, but so are some of the others.  Perhaps they can be shortened.

jrobinso:
I'm going to accept this PR, then make the changes suggested above.   Less complex than alternatives.

Issue #903, opened 2021-02-03
IGV won't start anymore on Mac


awilcz:
Hi, I was running IGV just fine until yesterday, when it started running away with CPU usage, and then after force quitting, it won't launch again. I've tried deleting all the relevant files, reinstalling etc and to no avail. I've tried launching it with Java 14 as well as Java 11... 

Here's my error message: 
`WARN [2021-02-03T12:24:08,520]  [Globals.java:135] [main]  Development mode is enabled
INFO [2021-02-03T12:24:08,536]  [DirectoryManager.java:180] [main]  IGV Directory: /Users/***/igv
INFO [2021-02-03T12:24:09,657]  [Main.java:205] [AWT-EventQueue-0]  Startup  IGV Version 2.9.0 02/01/2021 06:00 PM
INFO [2021-02-03T12:24:09,675]  [Main.java:206] [AWT-EventQueue-0]  Java 11.0.9 (build 11.0.9+11) 2020-10-20
INFO [2021-02-03T12:24:09,676]  [Main.java:209] [AWT-EventQueue-0]  Java Vendor: Oracle Corporation https://openjdk.java.net/
INFO [2021-02-03T12:24:09,677]  [Main.java:211] [AWT-EventQueue-0]  JVM: OpenJDK 64-Bit Server VM    
INFO [2021-02-03T12:24:09,849]  [Main.java:214] [AWT-EventQueue-0]  Default User Directory: /Users/***
INFO [2021-02-03T12:24:09,850]  [Main.java:215] [AWT-EventQueue-0]  OS: Mac OS X 10.15.5 x86_64
ERROR [2021-02-03T12:24:10,156]  [DefaultExceptionHandler.java:49] [AWT-EventQueue-0]  Unhandled exception
java.lang.NullPointerException: null
	at org.broad.igv.util.AmazonUtils.GetCognitoAWSCredentials(AmazonUtils.java:101) ~[igv.jar:?]
	at org.broad.igv.util.AmazonUtils.createPresignedURL(AmazonUtils.java:408) ~[igv.jar:?]
	at org.broad.igv.util.AmazonUtils.translateAmazonCloudURL(AmazonUtils.java:438) ~[igv.jar:?]
	at org.broad.igv.util.HttpUtils.mapURL(HttpUtils.java:146) ~[igv.jar:?]
	at org.broad.igv.util.HttpUtils.createURL(HttpUtils.java:135) ~[igv.jar:?]
[...]

PR #904, opened 2021-02-05
Fix NPE for indeterminate AWS tiers

brainstorm:
Hi @jrobinso,

While eyeballing the code for issue #903, I've noticed an AWS tier determination robustness issues that I'd to see merged as soon as you can & released, if you don't mind... perhaps the user could try it as a dev build first?

Also bumped to an AWS SDKv2 release that contains several bugfixes and bisected a `module-info.java` issue reported upstream to AWS:

https://github.com/aws/aws-sdk-java-v2/issues/2270

/cc @ohofmann @reisingerf

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,9 +131,10 @@ dependencies {
         [group: 'org.netbeans.external', name: 'AbsoluteLayout', version: 'RELEASE110'],
 
         // Amazon deps
-        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.15.9'],
-        [group: 'software.amazon.awssdk', name: 'sts', version: '2.15.9'],
-        [group: 'software.amazon.awssdk', name: 's3', version: '2.15.9']
+        [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.15.67'],
+        [group: 'software.amazon.awssdk', name: 'sts', version: '2.15.67'],
+        [group: 'software.amazon.awssdk', name: 's3', version: '2.15.67'],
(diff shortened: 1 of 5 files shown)


jrobinso:
Merged.   Is this a fix for a regression (i.e. a failure in 2.9 that would not happen in 2.8.13),  or a proactive fix for a new or unreported problem?  We will do a release next week regardless.

RE #903 I have some questions outstanding to the posters but no response so far.   There is something odd about the stacktrace, if I read it correctly the failure occurred from "s3" urls referenced from a user defined genome list file.   We have never supported s3:// urls for ".genome" or ".fasta" files, so its puzzling.

brainstorm:
Thanks! It's both actually, a user reported that some S3 URLs were not loading via the S3 File Chooser but they did when going through the "Load from URL" path plus also the fix should cover all corner cases (hopefully) if AWS shuffles the tier determination logic in the future.

W.r.t #903, yeah, without further user feedback I'm a bit unclear as what's going on. The only contentious code paths that might touch/affect those users might be related to `isAwsS3Path`?:

https://github.com/igvteam/igv/search?q=isAwsS3Path
https://github.com/igvteam/igv/blob/master/src/main/java/org/broad/igv/util/AmazonUtils.java#L447

But I do need a bit more context on the user's setup to be able to reproduce those, I couldn't reproduce modifying my existing setup.

jrobinso:
I'm planning to do a release on Wednesday, in the meantime the PR is merged to the "snapshot build" if anyone needs it quickly.

PR #917, opened 2021-02-22
Support FASTA files behind AWS S3 buckets

brainstorm:
Fixes issue #903 and allows users to define custom genomes hosted on their own S3 buckets (privately), namely:

```
% pwd
/Users/rvalls/dev/igv

% cat ~/igv/genomes/user-defined-genomes.txt
hg38.fa	s3://umccr-refdata-dev/genomes/hg38/hg38.fa	s3://umccr-refdata-dev/genomes/hg38/hg38.fa
```

I will update the UMCCR documentation accordingly if and when this PR gets merged.


jrobinso:
@brainstorm Just a heads up it will be a couple of weeks before I get to this.   I'm working on igv.js and igv-reports currently.  I do have one question, what happens when the signed URL to the fasta file times out?   With the other formats there is a refresh,  I don't see where that happens here.

I'm not suggesting we do a check and refresh in the way other formats are done.   I think signed URLs should be checked and fetched at a lower level,  in HttpUtils.openConnection in fact, where all other such things are checked.   When this is done all formats are supported automatically.    I think with that change, plus the improvements in this PR,  fastas and all else will be supported.    I will look into that when I return to IGV in a few weeks.

brainstorm:
@jrobinso No rush with this one, I was also off for a week. I have not tested the refreshing explicitly but since there's a checklogin on the code path is should re-request the auth flow to the user (AFAICT)?

I'm not sure about tackling all this at a lower level, I recall an unfruitful attempt in https://github.com/igvteam/igv/pull/620#discussion_r317592218 where it seemed a bit more involved than it looks? I'd be awesome if that's not the case and you can find a good solution that does not involve as much patchwork as the current status, for sure.

jrobinso:
@brainstorm I can't look at this now, but I'm pretty sure sequence (fasta) queries do not go through "check login".    I did try the low level fix once before, spent maybe an hour on it, and there were issues.   However done properly it should be less complex than the current flow.   Conceptually its very simple unless I'm missing something,  check if the URL being requested is expired, and if so refresh it.   When I have a block of time for this again I'll give it a proper go.    The main bit of this PR is still necessary to deal with startup timing.

brainstorm:
Thanks for the merge Jim! I hope you had some time to test the bits you were concerned about? 

In any case we'll have someone testing this for a while before deploying this change org-wide (after it gets released, that is).

jrobinso:
@brainstorm Stupid question, but where do I set the expiration time for presigned URLs?  Is it an S3 bucket policy?  I want to set it to something very short for testing.    

I'm currently working on support for all file types, by checking and regenerating urls at the lowest level, just before they are used, rather than instantiating new type-specific "readers".

jrobinso:
@brainstorm OK it looks like I can set this in Java, which is o.k. for testing purposes if I remember to set it back.    Perhaps expiration duration should be user settable, up to a maximum.

brainstorm:
Hey Jim,

As you might have noticed, most of the presigned URL expiring logic is within `AmazonUtils.java`. I'd be careful with a user settable expiration duration since the limits for that are very much dependant on the AWS side of things:

1) IAM users have 7 days expiration time when creating pre signed urls.
2) IAM **roles** have configurable 1-12h (max) settings. Non user-controllable, admin only.
3) AWS SSO, Cognito, etc.. (unless things change on some AWS re:Invent) should follow the regime on point 2.

So, in summary, even if the user sets a particular expiring time on the client side (IGV), AWS will either ignore or deny it, which can be a source of major confusion.

I'd do that test you are undergoing right now, set it back and let it be as it is, TBH ;)

PR #936, opened 2021-03-17
command -  sort by attribute

joshua-gould:


jrobinso:
@joshua-gould This will break if an attribute name just happens to equal an alignment sort option.  I think I would prefer an explicit command, since "sort" is taken. maybe "sortTracks" or "sortByAttribute".     This same critique could be applied to the seg copy # test, "region score", but in that case we have complete control of the vocabulary.

joshua-gould:
How about this syntax:
sortTracks trackName (asc|desc), trackName (asc,desc), etc...

jrobinso:
I don't understand that syntax,  what is trackName?  Sorting by attricture sorts tracks,  it doesn't make sense for an individual track.  Also,  what happened to the attribute argument?

Since this PR is to implement "sortByAttribute",   lets just call it "sortByAttribute".   That sorts all tracks, its what sorting by attribute does.   I assume the list defines secondary, tertiary, etc sorts.   So

sortByAttribute attribute1 (asc|desc),  attribute2 (asc|desc), ...

joshua-gould:
Yes, sorry I mean attribute instead of trackName. Should asc|desc be optional or mandatory?

joshua-gould:
I updated the command. Syntax is sortByAttribute attributeName asc|desc ....
Example: Sort by attribute1 descending: sortByAttribute attribute1 desc
Example: Sort by attribute1 ascending, then by attribute2 descending: sortByAttribute attribute1 asc attribute2 desc

jrobinso:
OK,  I'll look at this as soon as I can.

PR #937, opened 2021-03-20
Joshua gould master

jrobinso:


jrobinso:
I made a couple of minor changes,  supporting quoted arguments (for attribute names with spaces), and making attribute name specs case insensitve.   Unforunately I accidentally pushed unrelated stuff I was working on.  I don't see any way to undo this easily so I'm merging them together.

PR #938, opened 2021-03-23
Added commands fitTracks, showAttributes, showDataRange, setTrackHeight

joshua-gould:
Moved fitTracks to IGV in order to reuse code. Added showDataRange and attributes logic to CommandExecutor


jrobinso:
I won't be able to get to this right away, but had a quick look.    I think "attributes" is too generic, and its not a verb,  so maybe "showAttributes". 

I think this method belongs on TrackPanel (currently on IGV):  ``` private boolean fitTracksToPanel(DataPanelContainer dataPanel)``` and becomes ```fitTracksToPanel()``` or even just ```fitTracks()```.  No argument needed as it becomes just ```this.getScrollPane().getDataPanel()```.

PR #939, opened 2021-03-24
Added applicationHeaderPanel.getHeight() to snapshot height

joshua-gould:


jrobinso (changes requested):
There are a number of diffs showing for SnapshotUtilities.java,  but I think they are all just whitespace.   However the DEFAULT_MAX_PANEL_HEIGHT should not change, definitely should not be Integer.MAX_VALUE.

joshua-gould:
The bottom of the snapshot is cutoff without this change when creating a snapshot in batch. See the the two attached images for an example.
test_1000
test_max_value

jrobinso:
The bottom will always be cut off if you see a scroll bar, that's how its supposed to work.  If you want to see everything set maxPanelHeight to a large number in your script.    I do agree applicationHeaderPanel.getHeight() should probably be accounted for, that's not the change I commented on it was the other ones.

joshua-gould:
Just curious why you wouldn't want to save the image past the scroll bar?
Or does it make sense to add an option to save image to save the visible
tracks only?

jrobinso:
Well for 1 its been that way for 10+ years and we can't change the behavior for 10s of thousands of people.   The idea is you save a screenshot roughly approximating what you see.   Without a max height alignment screenshots could be 1/2 a million pixels deep or more.   But the basic answer is this would change existing behavior which we do not do.

joshua-gould:
Updated the pull request to only add applicationHeaderPanel.getHeight()

PR #940, opened 2021-03-24
Added window function absolute maximum.

joshua-gould:
Use case: For normalized values (e.g. z-scores) I want to summarize by the most extreme value. If a region has the values -5, 6, the summarized value is 6. If a region has the values -6, 5, the summarized value will be -6. In case, the negative and positive maximum values are tied, the positive value is selected.


jrobinso:
@joshua-gould Sorry for the long delay on this.    I think what you identified is actually a bug,  "max" should already use absolute values.    I can't imagine any use case where this would not be the case

PR #1008, opened 2021-07-30
S3 refactor

jrobinso:
Extend s3 authenticated access support to all file types.


jrobinso:
@brainstorm This PR should extend s3 support to all file types.    For the most part (tribble excepted, see below) we instantiate readers with s3 urls, rather than signed urls,  and generate/check signed urls at the point they are needed.

I had to dance around the htsjdk a little bit for "Tribble" formats,  bed, gff, etc,  as we must (currently) pass an https url to the htsjdk, it will not accept an s3 url.  

I will leave this PR open a few days in case you want to test it before I merge.

brainstorm:
Thanks @jrobinso, that's a good addition if all the other formats are now supported... I just gave it a go and doesn't seem to break in our environment, so welcome to merge right away if you wish? 👍🏻

jrobinso:
@brainstorm I tested it as well, I will merge and include it in the next release.    The tests are already fixed in the main branch, the failure was in the s3_refactor branch.

PR #1066, opened 2021-12-11
Upgrade log4j to 2.15.0 - CVE-2021-44228

rhowe:


jrobinso:
This has been done already.  Also, its not enough to just change the gradle settings 2.15 is not backward compatible with 1.2.

Having said that I don't think the version we were using (1.2) had this "feature", I think it came in 2.0 beta.  Furthermore we don't log user input text.  Nonetheless I have updated.

rhowe:
Good stuff. I thought the log4j-1.2-api library from log4j2 would continue to supply the log4j 1 API, but yeah, moving to log4j2 fully is even better.

jrobinso:
Actually I think you're right about that, there is a bridge, but the conversion was trivial.   I did test our existing IGV and it does not appear vulnerable to this current bug, but it was time to update anyway.   Eventually I might just switch to java.util.Logging.

PR #1099, opened 2022-02-05
Updated README.md and added more information into exon details.

generalbytes:
Added requirement for Java 11 fro gradle compilation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ we recommend the pre-built releases available at [http://software.broadinstitute
 
 Builds are executed from the IGV project directory.  Files will be created in the 'build' subdirectory.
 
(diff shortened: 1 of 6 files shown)


generalbytes:
I also added:
Details on exon now show also other important information such as: coding position within exon and gene as well as nucleotide position within gene, exon length and exon coding length.

jrobinso:
@generalbytes I'm not sure your coding position calculation is correct, at the very least it isn't self evident what you are trying to compute.   "genome coordinate" and "coding start" are in DNA coordinates, which of course include the introns.  At the very least we need more comment here.

getPosition should be named getRelativePosition if that's what it is.

Issue #1101, opened 2022-02-07
Unable to load from s3 URL


qing-at-work:
Tested versions: 2.12.0, 2.11.9

After configuring igv with Cognito authentication by following instructions [here](https://umccr.org/blog/igv-amazon/), loading from s3 URLs doesn't work (no error messages, no tracks loaded in IGV).

jrobinso:
Fixed with PR https://github.com/igvteam/igv/commit/104b3d97b0e9ded9ae1a18d7c5ba030a14c12917

PR #1103, opened 2022-02-07
Load tracks from s3 URL after accessibility check #1101

qing-at-work:
Load tracks from s3 URL after accessibility check #1101

### DIFF
--- a/src/main/java/org/broad/igv/ui/action/LoadFromURLMenuAction.java
+++ b/src/main/java/org/broad/igv/ui/action/LoadFromURLMenuAction.java
@@ -91,6 +91,10 @@ public class LoadFromURLMenuAction extends MenuAction {
 
                     final String url = mapURL(inputURL.trim());
 
+                    if (url.startsWith("s3://")) {
+                        checkAWSAccessbility(url);
+                    }
+
                     if (SessionReader.isSessionFile(url)) {
                         try {
                             LongRunningTask.submit(() -> this.igv.loadSession(url, null));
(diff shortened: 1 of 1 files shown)


jrobinso:
@qing-at-work Yes that is quite obviously a bug.  I suppose the Cognito users we have, which are mostly at UMCCR, were using the S3 "picker" to load files.

Looking at this, I think the s3 AWS check shoule be just before line 94, its possible someone stores a session file in S3.   Could you try placing it there, and if successful modify the PR?  This is what I think should be done.  

```
                   final String url = mapURL(inputURL.trim());

                    if (url.startsWith("s3://")) {
                        checkAWSAccessbility(url);
                    }
                    if (SessionReader.isSessionFile(url)) {
                        try {
                            LongRunningTask.submit(() -> this.igv.loadSession(url, null));
                        } catch (Exception ex) {
                            MessageUtils.showMessage("Error loading url: " + url + " (" + ex.toString() + ")");
                        }
                    } else {
                        ResourceLocator rl = new ResourceLocator(url.trim());

                        if (dlg.getIndexURL() != null) {
                            String indexUrl = dlg.getIndexURL().trim();

                            if (GoogleUtils.isGoogleCloud(indexUrl) || GoogleUtils.isGoogleDrive(indexUrl)) {
                                enableGoogleMenu();
                            }

                            rl.setIndexPath(indexUrl);
                        }
                        igv.loadTracks(Arrays.asList(rl));

                    }
```

jrobinso:
@qing-at-work Looks good, but could you confirm that you tested this in your environment?

qing-at-work:
> @qing-at-work Looks good, but could you confirm that you tested this in your environment?

Yes @jrobinso, I have tested this fix in my local environment (Windows Server 2019).

jrobinso:
OK thanks for the fix.

PR #1134, opened 2022-04-13
CommandExecutor.load should report errors to the client instead of throwing java.lang.Error

albertg1984:
IGV 2.10.0 has some changes to support data URLs. Some changes were necessary to the error handling and ever since certain error conditions result in a java.lang.Error being raised. This type of Throwable is never caught and will leave the client waiting for a
response indefinitely. As java.langError is reserved for unrecoverable errors the raise statements are replaced by return statements (similar to how the rest of the error handling is performed).


jrobinso:
Thanks for the PR

PR #1138, opened 2022-04-25
WIP: basic functionality if access keys are available

markotitel:
Hi,

I hope we can discuss on this PR based on comments here https://github.com/igvteam/igv/issues/1093

My background is AWS Architect Professional, Devops, Linux admin type of guy. I don't have Java experience. But I am in the business for 10+ years and "know my way around".

For this pull request I have hired Java dev proffesional to help me do a proof of concept.

**If any contributor is interested, I would very much like to pay for this feature/request**

I will try to describe use case for this feature.

Lets say a **Space Genomics** company has many bioinformaticians and reporters that are using IGV on a day to day basis.

- Reporters are interpreting data that is located in S3 buckets.
- Reporters are opening IGV app from a third party software which lists all the samples and provides "quick link" to open IGV with appropriate file location automatically (this already works)

Lets make assumption that S3 files are public. Based on this assumption IGV can open files using standard HTTPS protocol. IGV can be used sucessfully.

But what if **Space Genomics** company uses **private** S3 storage encrypted at rest?
We need to **sign** S3 urls before opening file location in IGV. This also is not such a big deal. We can make functionality in a third party application to sign the URL before opening.

But now reporter wants to share a session with someone. And the session can be opened after couple of days.
Signed URLs will expire. In addition to that, **Space Genomics** will sign URL based on **source ip**, in this case other reporters cannot open the session file since signed url won't work for them.

In **Space Genomics** company reporters log in to their workstations and Windows administrators automated workspace setup after logging in. Each time reporter logs into their account Windows will obtain and set system wide environment variables, among these variables we will have AWS variables set which adhere to AWS Java SDK requirements.

This way we can seamlessly enable reportes to open S3 urls without URL signing and without using Cognito > oAuth authentication/authorization.

In addition to this, we provide **Preferences** menu option to manually input acess_key/secret_access_key combination if a reporter wants or has custom aws keys which one can use to get the S3 object.

I hope this makes sense and is worth to check out and discuss on the pull request.

**EDIT:** I have to write additional notes.
Use case I described is fictional, but I work for bio-informatics company (1000+ employees) where this basic feature would be a life saver for multiple departments. Devops, Networking, Security and Geneticists/Reporters.

IGV should be able to load **both** HTTPS and S3 urls without going AWS Cognito route. AWS Cognito is a nice to have feature. But basic functionality to enable "native cloud api" access is "common sense" so to speak.

P.S. No harsh language or tone intended, I am not a native English speaker especially not a writer :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,7 +123,9 @@ dependencies {
             // Amazon deps
             [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: '2.16.7'],
             [group: 'software.amazon.awssdk', name: 'sts', version: '2.16.7'],
-            [group: 'software.amazon.awssdk', name: 's3', version: '2.16.7']
+            [group: 'software.amazon.awssdk', name: 's3', version: '2.16.7'],
+            [group: 'software.amazon.awssdk', name: 'auth', version: '2.16.7'],
+            [group: 'software.amazon.awssdk', name: 'profiles', version: '2.16.7']
     )
 
     testImplementation(
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -36,4 +36,5 @@ module org.igv {
     requires software.amazon.awssdk.http;
     requires software.amazon.awssdk.utils;
     requires com.fasterxml.jackson.core;
+    requires software.amazon.awssdk.profiles;
 }
--- a/src/main/java/org/broad/igv/prefs/IGVPreferences.java
+++ b/src/main/java/org/broad/igv/prefs/IGVPreferences.java
@@ -53,12 +53,15 @@ import java.io.PrintWriter;
 import java.util.*;
 
 import static org.broad.igv.prefs.Constants.*;
+import static software.amazon.awssdk.profiles.ProfileProperty.AWS_ACCESS_KEY_ID;
+import static software.amazon.awssdk.profiles.ProfileProperty.AWS_SECRET_ACCESS_KEY;
 
 /**
  * Manages user preferences.
  */
 public class IGVPreferences {
 
+    private static final String AWS_REGION = "aws_region";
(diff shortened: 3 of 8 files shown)


brainstorm:
> This way we can seamlessly enable reportes to open S3 urls without URL signing and without using Cognito > oAuth authentication/authorization.

Why do you find AWS Cognito complex to deploy in your Space Genomics org?

jrobinso:
@brainstorm @markotitel This is not really the forum for a best practices on AWS debate.   @brainstorm 's point about this solution not clashing with the existing Icognito implementation is a good one.   In what way could they clash?  For users who have keys defined in the magic places, hereafter referred to as magic keys,  AND are using Incognito for access?  Let's think about potential clashes.    If users have Incognito setup for IGV as documented elsewhere perhaps we just ignore magic keys.

There are a more changes in this PR than I would expect to be necessary.  According to the AWS JDK documentation the JDK should be picking up credentials according to the order specified in the credentials chain automatically,  the question is why is it not?     Perhaps its the way we are creating an S3Client.

jrobinso:
@markotitel Please give me a few days or even a week to get up to speed on this before doing further work.    I think we can come to a solution.

markotitel:
I would like to thank you for taking time to review this "idea" :)

I really do think that this can help a lot of people in the bioinf world.

jrobinso:
OK I have an alternative solution working in principle, it will take another day or so to cleanup and test.   The basic idea is if cognito is configured, those credentials are used.   If not default credentials are used.  IGV itself doesn't know anything about the credentials per se.   The "s3" url is saved to the session.

Leave this PR open in case I have to refer back here for anything.

jrobinso:
I'm going to close this PR in favor of #1139 .  Let's discuss that solution there, if needed.

PR #1139, opened 2022-04-27
Aws default credentials

jrobinso:


jrobinso:
This PR enables loading from AWS S3 using s3:// urls if Amazon credentials can be found in the default credential chain as described here:  https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html

In addition,  if default credentials are found,  the default region chain will be searched as described here: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html.   Its not clear that Region is critical as we are not doing any uploads (puts),  but it is required in some API calls.

Existing Cognito users should be unaffected.   Cognito configuration, as described https://umccr.org/blog/igv-amazon-frontend-setup/)will take precedence over default credentials.   To be clear, if front end provisioning with Cognito is setup as described at the UMCCR site login will be required and no search will be made for default credentials.

markotitel:
Hi,
I've just built the code and did some tests.
Quick tests were successfull!! 🚀 🚀 
How did I test:
- In PowerShell set environment vars
$Env:AWS_ACCESS_KEY_ID=
$Env:AWS_SECRET_ACCESS_KEY=
$Env:AWS_REGION=
- Start IGV from the same shell to preserve environment vars.
- Amazon menu is displayed and I can list the buckets
**But more importantly**
- File > Open URL > S3 URI works !! 🚀 

Let me know what other testing I can do (if any).

markotitel:
Okay, I've done some more testing.

I will do my best to describe the next test case.

When a user logs in to Windows workstation using Microsoft AD credentials, the login sesion automatically obtains AWS SSO credentials with `aws sso login`.
This will generate Temporary AWS Credentials which includes.
- $Env:AWS_ACCESS_KEY_ID=
- $Env:AWS_SECRET_ACCESS_KEY=
- $Env:AWS_SESSION_TOKEN=

After these vars are set automatically on login, IGV can successfully load the S3 file.

Our policy is that these temporary credentials are refreshed every hour and new $Env values are set behind the scene.

This is where IGV is still using old temporary credentials and is unable to load the file because we refreshed the temporary credentials.

To overcome this a user has to close IGV and open again. It is fine since users mostly do not keep IGV open for an hour while working. But in some cases it can happen.

I hope description makes sense.

jrobinso:
@markotitel  This might be difficult to overcome as it looks like those credentials are cached in the AWS JDK.    Did this case work with the (now closed) PR #1138?

markotitel:
Hi, no this was not working. Yes I know this is hard to do. Just wanted to mention and to show that I did intensive testing 😃 

Other than that reported "issue", this feature is working perfect I can say.
I have already showed to couple of colleagues and reaction was more than what I excpected.

jrobinso:
Yes appreciate the testing.  I also tested all combinations of default credentials + cognito + nothing.   This will be in the next release,  estimate middle of May.

jrobinso:
@markotitel this is now released in version 2.13.0

Issue #1130, opened 2022-04-02
Support bigGenePred


jrobinso:
Support bigGenePred, used for T2T annotations.   This is not genePred, the column layout is closer to bed.  See https://genome.ucsc.edu/goldenPath/help/bigGenePred.html.

Also bigNarrowPeak and bigInteraction

jrobinso:
@maximilianh Yes this looks good, its at the top of my list along with assembly & track hub parsing.  BTW are the NCBI hubs you speak of the "Gen Ark" hubs (https://genome-blog.soe.ucsc.edu/blog/2021/11/23/genark-hubs-part-1/)?    I find the public hub page a little overwhelming, all kinds of stufff in there (this page. https://genome.ucsc.edu/cgi-bin/hgHubConnect)

maximilianh:
The public hubs page is something different. Public hubs have been
submitted by all sorts of labs. We will need to clean up this page and add
hubs groups and/or faceting so people can find hubs again.

The Genark hubs (so currently only NCBI assemblies) are linked from this
page:
https://hgdownload.cse.ucsc.edu/hubs/

(We're preparing a big index page that shows all of them.)

Also, you can find any hub by searching on our main page in the search box
http://genome.ucsc.edu/cgi-bin/hgGateway e.g. you can type "human" there
and find CHM13.
(this page does not make it sufficiently clear that you can search for much
more than the UCSC-genomes... sigh...)

jrobinso:
@maximilianh Its not obvious how to get from the hub URLs, which link to a browser page, to the trackDB.txt and related files,  for example  https://hgdownload.soe.ucsc.edu/hubs/GCA/009/914/755/GCA_009914755.4/trackDb.txt

Maybe I have missed a link but can't find one.    I don't even remember how I managed to to find the link above, some combination of web searching and guessing.   Thee seems to be another trackDb.txt for that assembly (chm13 v2.0) here, but with many fewer tracks:  https://t2t.gi.ucsc.edu/chm13/hub/t2t-chm13-v2.0/trackDb.txt

WRT bigGenePred and other specialized "bb" formats there is an issue determining format if not given external information.   I had hoped "autoSql.table" would tell me the format but it doesn't.   For example the following stanza indicates a type, which I'm interpreting as format, of "bigGenePred",  however the actual structure differs from what's described for that format and "autoSql.table" indicates the format is "bigCat".   I supposed I can parse the autoSql, which I'm already doing, and infer a format from that.

```
[...]

PR #1140, opened 2022-05-04
Bb

jrobinso:


jrobinso:
Support for bigGenePred and bigNarrowPeak formats - fixes #1130

PR #1144, opened 2022-05-25
Bigbed

jrobinso:


jrobinso:
Enhancements for bigbed parsing - parse all extra fields and store as attributes.  Compute exon frames for any bb file with field named "exonFrame".   See issue #1130

PR #1154, opened 2022-06-12
option in preferences to enable paired end mode by default

ssadedin:
This change adds an option in preferences so that alignments are loaded with the "View as pairs" option enabled by default.

While it's a minor change, it's a huge a time saver when you have a lot of different files to review, and the paired end signal is important (eg: structural variants). Before this change, I have to laboriously right click and select "View as pairs" on each of my alignment tracks before I can start. With this change, I can enable that as the default in settings so that all files are loaded that way by default.

Since there are only two options, I just made this a checkbox:

image


jrobinso:
Looks good, thanks.

PR #1157, opened 2022-06-14
Update to htsjdk 3.0.0 which includes support for CRAM reference regions

lbergelson:
@jrobinso This removes the long wait when loading a cram contig initially.  I thought it added slight scrolling delay when jumping along the contig but with further testing I'm not seeing it anymore and the profiler doesn't show anything significant.  I think my first testing was using different zoom levels by accident so I was seeing read loading lag on one and not the other.  

I removed all the complex code associated with loading the full contig under the assumption that htsjdk is never going to call them method in the CRAM use case.  Does IGV ever write cram files?  

I could replace the original with an UnsupportedOperationException instead so we would know if it ever gets called by accident, or keep the complex caching logic in place, although that seems bad.

### DIFF
--- a/src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
+++ b/src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
@@ -50,120 +50,23 @@ public class IGVReferenceSource implements CRAMReferenceSource {
 
     private static Logger log = LogManager.getLogger(IGVReferenceSource.class);
 
-    static ObjectCache<String, byte[]> cachedSequences = new ObjectCache<>(5);
-
-    static GenomeChangeListener genomeChangeListener;
-
-    static HashMap<String, Object> locks = new HashMap<>();
-
     @Override
     public byte[] getReferenceBases(SAMSequenceRecord record, boolean tryNameVariants) {
+        return getReferenceBasesByRegion(record, 0,  record.getSequenceLength());
+    }
 
-        final String name = record.getSequenceName();
+    @Override
+    public byte[] getReferenceBasesByRegion(final SAMSequenceRecord sequenceRecord, final int zeroBasedStart, final int requestedRegionLength) {
+        final String name = sequenceRecord.getSequenceName();
         final Genome currentGenome = GenomeManager.getInstance().getCurrentGenome();
         String chrName = currentGenome.getCanonicalChrName(name);
-        Chromosome chromosome = currentGenome.getChromosome(chrName);
+        byte[] bases = currentGenome.getSequence(chrName, zeroBasedStart, zeroBasedStart + requestedRegionLength, true);
 
-        byte[] bases = cachedSequences.get(chrName);
-
-        if (bases == null) {
-            try {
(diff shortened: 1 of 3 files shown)


src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
```diff
@@ -50,120 +50,23 @@ public class IGVReferenceSource implements CRAMReferenceSource {
...
-                synchronized (lock) {
-
-                    if (IGV.hasInstance()) IGV.getInstance().setStatusBarMessage("Loading sequence");
-                    bases = currentGenome.getSequence(chrName, 0, chromosome.getLength(), false);
-
-                    // CRAM spec requires upper case
-                    for (int i = 0; i < bases.length; i++) {
-                        if (bases[i] >= 97) bases[i] -= 32;
-                    }
+        byte[] bases = currentGenome.getSequence(chrName, zeroBasedStart, zeroBasedStart + requestedRegionLength, true);
```
lbergelson, line 63: I'm operating under the impression that Genome uses 0-based half open coordinates.  Let me know if that's incorrect.

src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
```diff
@@ -50,120 +50,23 @@ public class IGVReferenceSource implements CRAMReferenceSource {
...
 
-    static ObjectCache<String, byte[]> cachedSequences = new ObjectCache<>(5);
-
-    static GenomeChangeListener genomeChangeListener;
-
-    static HashMap<String, Object> locks = new HashMap<>();
-
     @Override
     public byte[] getReferenceBases(SAMSequenceRecord record, boolean tryNameVariants) {
+        return getReferenceBasesByRegion(record, 0,  record.getSequenceLength());
```
lbergelson, line 55: In the previous implementation there was some extra stuff done to get the length of the chromosome from the Genome object instead of the SAMSequenceRecord.  I changed that because it seemed that if they were different then there's already a bug somewhere since the CRAM file is expecting this to be the matching reference.  Happy to add it back if I'm missing some detail.

src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
jrobinso: That's correct, 0-based 1/2 open is used internally for all objects.    Conversion is done when parsing 1-based file formats, or presenting or accepting coordinates from the user.

src/main/java/org/broad/igv/sam/cram/IGVReferenceSource.java
jrobinso: The chromosome name will not neccessarily match the chromosome name in the CRAM file,  for example the IGV reference might use "chr1" and cram file "1".   The sequence must match of course but chromosome names do not.  This might be what the "extra stuff" was about,  I will look at it.

jrobinso (approved):
Looks good, big improvement.

Issue #1117, opened 2022-02-24
Feature request: sort by read length


gzentner:
We look at a lot of Nanopore direct RNA-seq and have found it nice to look at reads stacked in descending order of length. For positively stranded reads, this is easily accomplished by sorting by start coordinate; however, for negatively stranded reads the end coordinate in the BAM is the true start coordinate of the read and so the sorting gets wonky as a result. It would be great if there was an option to either sort by read length or if the sort by start option was strand-aware.

jrobinso:
Sort by read length is a reasonable option,  but to clarify do you really mean read length or aligned length?  i.e. length of the read sequence or (end - start) of the alignment?

gzentner:
Aligned length, apologies. Since these are DRS reads they can be quite long so the aligned length gets up there. The CIGAR strings are often a mess because Nanopore is so noisy, as well. Thanks!

gzentner:
Thinking about this more, because the DRS reads are single-end by nature, I'm guessing that the CIGAR string would need to be used for this, in conjunction with the leftmost aligned base column.

akkadadam:
Will this feature be integrated into a future update? If so, is there a ballpark estimate of when that would be? Thank you very much for making this a reality!

jrobinso:
It will be in the next release, late July or early August.

PR #1159, opened 2022-06-17
Sort by Aligned Read Length

lbergelson:
* Add a new sorting option to allow sorting by the aligned read length.
* closes https://github.com/igvteam/igv/issues/1117


jrobinso (commented):
Looks good.

PR #1160, opened 2022-06-22
Enable other Oauth providers work with IGV

tejaswi4000:
This PR enables other oauth providers to load bam files, session files into IGV by using custom json file which has the client_id, client_secret, authorization_endpoint, auth_provider details. An example of config json can be found here: oauth-config-2.13.0.json


jrobinso:
@tejaswi4000 Sorry for the delay I've been on vacation.  It might take me a few days to get to this, but could you comment on what motivated the changes, other than what's in the title?  This might become obvious during the review, but I know some for example use Microsoft/Azure with the current implementation.   I have no way to test that personally.

For that matter I have no obvious way to test this PR, if you have suggestions for how to do that please let me know.

tejaswi4000:
@jrobinso No problem. The motivation behind these implementation was, we have most of our sample files (bam files and session files) stored in our internal server which requires oauth. So to access those files we required to patch the current igv version. We believe that this functionality can be used by other clients as well which requires oauth. From our end, we have tested it using a config json (attached in the repo oauth-config-2.13.0.json) which has the token and secret keys specific to our organization.

jrobinso:
This looks reasonable, I will try to get to it by the end of this week.

tejaswi4000:
Sounds good. Thank you!!

tejaswi4000:
Hi @jrobinso Is there anything I have to provide to get this PR approved?

jrobinso:
@tejaswi4000 No it looks good, sorry I've been delayed but other priorities but I will get to this soon.    Is there a reason you closed it?

tejaswi4000:
Cool. Not sure when I have closed. May I might have clicked on comment and close accidently. Thanks for reopening.

jrobinso:
@tejaswi4000 You added a config file with endpoints and keys, etc, for the Mayo Clinic.  Is that intentional?

tejaswi4000:
@jrobinso Yes Rob, This config file will be used if they were to use the oauth for their organization.

jrobinso:
@tejaswi4000 This config file will only work for the Mayo, and it looks like sensitive information is exposed here.  I hesitate to expose this.

```
{
  "client_id": "hbq82djj-wxky-7iub-j7zq-7i8nv72n48nq",
  "client_secret": "cVAX64fXRikCLmtLEb/cktrAtaHz/tmB3WHKhnbXN2Gq",
  "authorization_endpoint": "https://login.microsoftonline.com/77nwe2q2-e53r-uq2p-7vdh-9z7px83zmtiv/oauth2/authorize",
  "token_endpoint": "https://login.microsoftonline.com/77nwe2q2-e53r-uq2p-7vdh-9z7px83zmtiv/oauth2/token",
  "hosts": ["prod_host.mayo.edu", "dev_host.mayo.edu", "int_host.mayo.edu"],
  "auth_provider": "Mayo Clinic",
  "app_id_uri": "https://orgtools.onmicrosoft.com/6q9qk3mr-tw99-eu73-vx5a-nqw2aqidutm9",
  "scope": "openid",
  "find_string": "dept",
  "replace_string": "dept-oauth2"
}
```

tejaswi4000:
The tokens and urls are already masked and the values are not actual Mayo values but faked ones.

jrobinso:
OK so the purpose is an example?  Could you write up a short readme text on how to use this new capability?  Its not obvious.  My biggest issue right now is probably testing it, any suggestions?

tejaswi4000:
@jrobinso Would it be possible to meet for 30-minutes to show you how the oauth patch works for files on a server with oauth enabled?

jrobinso:
I don't think we need to meet, I understand how oAuth works,  what I'm asking for is a short writeup explaining to an end user how to use this new option.   Its not appropriate to put oauthConfig-custom.json in the root of our repository,  I am assuming that is an example file which end users would change and store somewhere on their local machine, as we do for google and amazon oAuth setup.   In any event a set of written instructions on configuring a custom oAuth server is what I am requesting.

Thanks

jrobinso:
BTW we can't arbitrarily change the name of custom configs from "oauth-config.json" to "oauth-config-custom.js",  this would obviously break current users custom oAuth configurations (line 112 of OauthUtils).  Are you aware of other potentially breaking changes?

jrobinso:
Also of concern is the check or URLs when loading a session for possible authentication.  This should not be necessary.  For Google protected files this is done reactively, in HttpUtils around line 877.    This handles all cases, whether from session or a user loading a file from URL,  and only prompts login if needed.   Could this strategy be used?


```
                } else if (code == 401) {
                    if (GoogleUtils.isGoogleURL(url.toExternalForm()) && retries == 0) {
                        GoogleUtils.checkLogin();
                        return openConnection(url, requestProperties, method, redirectCount, ++retries);
                    }
```

jrobinso:
Two more questions then I will wait for a response. 

* Why do we need the method ```getInstanceNoCreate()``` on OauthUtils?

* I do not have a way to test this, and we don't usually release features we can't test.   Are you aware of any test servers that could be used?  I might be able to use Google as a simulated test server, but that's not ideal.   This is not a show-stopper but is a concern.
*

jrobinso:
I'm leaving Saturday for a 10 day vacation,  between now and then I'm gong to make some changes and push them to a branch for your review.   What's here at the moment would break the application for existing users.    If I could get answers to questions above it would be helpful, otherwise I will make guesses.   Unfortunately I am unable to test against Azure.

One of the unknowns for me is what does Azure return if an attempt is made to access a protected file without an authorization token.    Google returns a 401 response code.   There's an assumption in this PR that all files on a listed host are protected, this is not necessarily the case, there are many public files on Google servers.    That is why we don't force a login based on the host.

jrobinso:
@tejaswi4000 I've made some changes and pushed to the branch https://github.com/igvteam/igv/tree/Tejaswi178-oauth_integration.   Here's a summary

* Change of name from oauth-config.json to oauth-config-custom.json rolled back, this would be a breaking change
* Assumed "appliesToUrl" for Google urls rolled back,  this is often not the case.  We use many Google files that are public and do not require login.
* Potential OAuth login is checked for every http request in HttpUtils.   This removes the need to check in the load-from-url and load-session actions.  

These changes address the concerns above and moves this close to merge status.   Could you test this branch in your environment?  You will need to rename your oauth-config-custom.json file to oauth-config.json, and place it in the IGV directory.

Thanks

Jim

jrobinso:
@tejaswi4000 Pinging as I'd like to resolve this in the next week or so.   Do you have any feedback on the changes proposed above?

tejaswi4000:
Hi Jim, sorry for the delay in testing.
I was able to test the changes made from your Merge. One concern would be with changing the name of oauth config json from its custom name is user might not be able to use 2 providers at the same time and we are aware of that. But other than that everything looks reasonable and working in our environment.

jrobinso:
@tejaswi4000 Thanks for verifying.

Good point re multiple providers,  that should be addressed, its a longstanding issue.   Currently users can't configure Amazon + Google for example.   Using a "-custom" would allow at least 1 additional provider.   I will look at that again.   The code in "http-utils" would need some logic to determine which oAuth provider to use with which host.   Ideally we should allow multiple provider configurations.

Its a good point you raise.

jrobinso:
@tejaswi4000 I've merged to an intermediate branch, next up merging my changes and testing.    Also, BTW, there might be some misunderstanding in the purpose of "oauth-config.json".   It has always been there as a means to customize oAuth, so adding "custom" to the name does not change anything or add any additional flexibility.   There are some who use oauth-config.json now for customizing oAuth, and changing the name would be a breaking change for them.   The issue of not being able to specify more than 1 oAuth provider is an orthogonal issue.

tejaswi4000:
Awesome. Thanks for merging the changes to master. I agree with you that the issue of not being able to specify more than 1 oauth provider is a different issue than this one.

jrobinso:
@tejaswi4000 This will be released in 2.14.2 ~ the end of the month.  In writing up documentation I had a question about "findString / replaceString".   I can see from the code what these do, but what does this have to do with OAuth?   I tried googling but didn't find much.   Is this a common thing to do when using Azure?

dedaimia:
@jrobinso I I'm another developer who works with @tejaswi4000
With regards to the findString / replaceString it's to support websites and apis that support multiple authentication methods with slightly different urls.

The main website @tejaswi4000 is testing against for example will normally prompt the user for an interactive login process, then use session tokens thereafter.  If you find a public cytobands file interactively on the website, the url will be https://<base url>/public/data/refdata/20201103/hg38/cytoBand_assembled.json

Unfortunately, the website admins won't accept a bearer token for login on that same url.  To get the same file with a bearer token you have to call https://<base url>/**oauth2**/public/data/refdata/20201103/hg38/cytoBand_assembled.json

It's extremely un-user friendly to tell them they have to manually change the url of every single file before they can open in IGV, so we put in the search and replace to switch the interactive url for the oauth2 bearer token one.

jrobinso:
OK, thanks for the explanation!

PR #1163, opened 2022-07-06
Methylation view

jrobinso:


jrobinso:
* Implements 5mC base modification methylation view -- fixes issue #1133 
* Improves coverage track for general base modification (MM tag) view
    * Height of colored bar is now proportional to average likelihood of modification over all called bases
    * Likelihood threshold used to determine if bar is colored or not -- analogous to "snp" (standard) view.  Defaults to 0.

Issue #1112, opened 2022-02-16
Add option to group by mapping quality (alignment tracks)


jrobinso:
Add "groupBy" option.  Split off from #1110.

lbergelson:
closed by #1165

PR #1165, opened 2022-07-08
Add options to group and shade by mapping quality

### LINKED ISSUE #1110: Feature request: color by mapping quality
It would be useful to have both implemented when looking at low complexity regions and trying to identify mapping artifacts.

lbergelson:
* add a new grouping option to group by mapping quality
* added a new "shading" submenu to allow shading alignments by mapping quality.
  Shading can highlight either high or low mapping qualities.
  New configuration options allow setting the high/low map quality categories, they default to 60/0
* closes #1110 and #1112

@jrobinso I kept the 3 new options in preferences, if you think they're too confusing I can remove them, but it seemed much harder to add a new dialogue box and people might want to be able to configure the color thresholds (most bams I know max out at 60 mq but the range goes up to 255 so I expect there are different ones that go much higher).
 
I think I've modified all the various places that need modifying but it's very possible I've missed somewhere.  

Screen Shot 2022-07-08 at 6 34 05 PM


jrobinso:
Looks good.   What happens if "shade by mapping quality" is on to the "zero MQ alignments" which are curently flagged with transparency and an outline?   I'm not sure what should happen, maybe nothing,  zero has a special (though technically undefined) meaning, its not really a possible value for mq.

jrobinso:
All set now,  I made a small change to treat "flag mq == 0" and "shade" as separate options, the "flag" takes precedence when its enable and mq == 0.   These preferences are grouped together in the UI.

Issue #1168, opened 2022-07-12
Incorrect URLs in server requests when trying to open TDF files


ubuntolog:
Hello everyone,

we have our own server that requires a secure token to access a file. URLs look roughly like that:

https://ourserver.com/filename.bam?token=some_text

When IGV tries to access the files from our server, it looks for index files at the same location. E.g. for BAM files it will add BAI extention to the filename and the URL will look like that:

https://ourserver.com/filename.bam.bai?token=some_text

I have been checking the server logs recently and found some failed attempts from IGV to access TDF files: it was not able to pass the token check. I did some debugging and found out that for TDF files IGV overwrites the URLs incorrectly. It adds the TDF extension to the token, but not to the filename that proceeds the parameters. Roughly like that:

https://ourserver.com/filename.bam?token=some_texttdf

The server thinks the token is invalid and returns the corresponding error code. The issue can be reproduced with the version 2.13.1. It is by no means a critical problem. I just wanted to let you know that it exists. Could someone please look into it?

jrobinso:
@ubuntolog Thanks for the report.   It will still not find the "tdf" files in most cases, of course, because not many use this option, but at least it won't find it for the correct reason.  Its from very early IGV,  if a "tdf" is discovered it uses it for coverage data.

PR #1169, opened 2022-07-12
Fix searching for TDF file when url has a query string

lbergelson:
* Correctly append ".tdf" to a url when there is a query string
* Fixes https://github.com/igvteam/igv/issues/1168

@jrobinso I'm pretty sure this fixes the problem but I'm not sure how to test it.


jrobinso:
Thanks @lbergelson it looks correct but I will test it then merge.

jrobinso:
Yep, it does the right thing

ubuntolog:
Thank you

Issue #1158, opened 2022-06-17
Sort orders / groupings should toggle between ascending/descending each time you select them


droazen:

jrobinso:
Resolved with PR #1180

PR #1180, opened 2022-08-09
Adding an option to reverse the order when sorting.

lbergelson:
* Adding a new option to the sort menu which reverse Alignment sorting when enabled
* Some of the sorting methods were moved and restructured.
* An alternative to #1158

Sorting behavior should remain unchanged if the new option is not activated.
Selecting or unselecting "reverse sort" will trigger a new sort operation at the currently selected location.

2 new fields are saved into the session files, I'm not sure if this is the right approach.  Currently sorting is treated as a global action, initiating a sort sorts all tracks at the same time in the same way. Adding a per-track option to the session file potentially allows manually creating sessions with inconsistent sorting, although I don't think that's a problem in any practical way and might be useful in some extremely rare circumstances.  

Maybe there should be a different place to persist this sort of information.

Sorting is also not persisted in a session file, it might be a good idea to enable that although it seems a bit tricky.

### DIFF
--- a/src/main/java/org/broad/igv/batch/CommandExecutor.java
+++ b/src/main/java/org/broad/igv/batch/CommandExecutor.java
@@ -829,7 +829,7 @@ public class CommandExecutor {
 
         if (sort != null) {
             final AlignmentTrack.SortOption sortOption = getAlignmentSortOption(sort);
-            igv.sortAlignmentTracks(sortOption, sortTag);
+            igv.sortAlignmentTracks(sortOption, sortTag, false);
         }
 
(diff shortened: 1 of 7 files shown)


lbergelson:
An example of the very questionably useful "reverse sort by bases".  Others should be more useful.

Screen Shot 2022-08-09 at 5 47 09 PM

It's very literal about reversing the sort though, so you end up getting all the reads which don't actually overlap that site at the top.  Maybe nulls should be treated specially and always sorted at the bottom even when reversing it?

PR #1188, opened 2022-08-13
Add a new color by Read Order option

lbergelson:
* This colors reads by first in pair and second in pair.


lbergelson:
I noticed this was missing when I was doing some changes to sorting and was trying to test "Sort by Read Order"

jrobinso (approved):
Looks good

PR #1189, opened 2022-08-13
Refactoring and Improving Sort

lbergelson:
* Several sorts were also improved in the process.
  - sort by base now includes insertions and deletions more correctly
    - insertions are included if they are inserted directly before or directly after the current base
    - deletions now are subsorted by length and starting position
    - several sorts that were previously done by hashcode are now lexicographical instead which
      is hopefully better in most cases
* Invert Sort now works much better and doesn't sort empty lines on top
* Big refactoring of the internal sorting code to make it Comparator driven
* Sort orders are now part of the SortOption enum which has been moved be be a toplevel class


lbergelson:
I ended up finishing this because it seemed like the sanest way to fix the weird empty lines at the top when inverting the sort.  There are still some todos/ things to revisit.

1.  Is sort by base working as well as it should be.  I had it include insertions that are either to the right or left of the selected 
position but it might be better to only include ones on one side.

It usually looks pretty good:
deletion sorting mixed

There's a bit of weirdness when there is an indel adjacent to a SNP though, not sure what the ideal solution is.
Screen Shot 2022-08-13 at 5 37 26 PM

2. Is there a way we can make tags sort better than hashcode does
3. It would be good to add a comparator that understand chr20 comes after chr3 because it's very annoying the way things work currently (and before this change.)
4. I should move the human readable names to be part of the enum so they don't have to be written into both AlignmentTrack and CommandLineExecutor.

PR #1195, opened 2022-08-16
JPasswordField (instead of JTextField) for password in proxy settings

ubuntolog:
Currently in the Preferences the proxy password is displayed as a plain text (however, it is stored in the file in an encoded form, which is good). To me it looks like a security issue, since someone may accidentally see the password. This is why I am creating this PR. Please have a look.


jrobinso:
Thanks

PR #1196, opened 2022-08-16
Add feature threshold indicator to zoom widget

lbergelson:
@jrobinso 
This is a minor feature that I have wanted in the past so I figured I would try implementing it.   Good learning project to play with some of the stuff I haven't touched yet.

- implementing getVisibilityWindow() on coverage and alignment tracks
- Adding an indicator to the Zoom control to show which zoom levels will display reads / features

This is what it looks like when there are 2 different thresholds set (variants AND reads) notice the two shades to indicate the separate thresholds.

Screen Shot 2022-08-16 at 3 33 45 PM

I implemented it in a kind of hacky way so I'd really like to hear feedback about better ways to get the threshold from the tracks into the widget without poling through them each time you redraw.  It's probably not that expensive so it may not be worth doing anything smarter.

Also happy to change the look of it.

### DIFF
--- a/src/main/java/org/broad/igv/ui/panel/ZoomSliderPanel.java
+++ b/src/main/java/org/broad/igv/ui/panel/ZoomSliderPanel.java
@@ -54,7 +54,7 @@ import java.util.stream.Collectors;
 public class ZoomSliderPanel extends JPanel {
 
     private static final Color TRANSPARENT_GRAY = new Color(200, 200, 200, 150);
-    private static final Color TRANSPARENT_BLUE = new Color(27, 96, 246, 50);
+    private static final Color TRANSPARENT_BLUE = new Color(27, 96, 246, 25);
     static Color TICK_GRAY = new Color(90, 90, 90);
     static Color TICK_BLUE = new Color(25, 50, 200);
 
--- a/src/main/java/org/broad/igv/ui/panel/ZoomSliderPanel.java
+++ b/src/main/java/org/broad/igv/ui/panel/ZoomSliderPanel.java
@@ -33,13 +33,20 @@
  */
 package org.broad.igv.ui.panel;
 
+import org.broad.igv.track.Track;
+import org.broad.igv.ui.IGV;
 import org.broad.igv.ui.util.IGVMouseInputAdapter;
 import org.broad.igv.ui.util.IconFactory;
 
 import javax.swing.*;
 import javax.swing.event.MouseInputAdapter;
+
+import java.util.List;
 import java.awt.*;
 import java.awt.event.MouseEvent;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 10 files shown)


jrobinso:
This is a neat feature, however we can have many tracks loaded each with a unique visibility window.    Think of 10 or 20, a combination of VCF, bigBed, and indexed annotations.    This is not uncommon as you might think.

src/main/java/org/broad/igv/ui/panel/ZoomSliderPanel.java
```diff
@@ -193,9 +200,33 @@ protected void paintHorizontal(Graphics g) {
...
+                    .map(Track::getVisibilityWindow)
+                    .filter(i -> i > 0)
+                    .sorted()
+                    .distinct()
+                    .map(threshold -> this.getReferenceFrame().calculateZoom(0, threshold))
+                    .collect(Collectors.toList());
+
+            transGraphics.setColor(TRANSPARENT_BLUE);
+            Rectangle maxZoom = zoomLevelRects[zoomLevelRects.length - 1];
+            for (Integer window : visibilityThresholds) {
```
jrobinso, line 221: There's a bug here somewhere,  array index can be negative.   I'll drop more details in the "conversation" widget

jrobinso:
@lbergelson Looks good overall, but there is a bug.    I'm not sure how I got in this state, but here are the previous steps

(1) load the session test/sessions/base_mods/HiFi_basemodfications.xml from the repository (File > Load Session...)
(2) From "Load from server" loaded Common Snps 1.4.2 from "Annotations - Variations and Repeats", and "Gencode v4.0" from "Annotations - Genes".
(3) Set the visibility window of the Snps track to 1,   visibility window of Gencode track to -1.
(4) Tried to remove the alignment & coverage tracks to see effect on zoom bar, at this point I noticed it was frozen and stack trace.   I'm not sure when the stack trace was created, it could have been prior to (4)




```
SEVERE [Jan 10,2023 09:53] [DefaultExceptionHandler] java.lang.ArrayIndexOutOfBoundsException: Index -3 out of bounds for length 22
	at org.broad.igv.ui.panel.ZoomSliderPanel.paintVisibilityThresholds(ZoomSliderPanel.java:222)
	at org.broad.igv.ui.panel.ZoomSliderPanel.paintHorizontal(ZoomSliderPanel.java:204)
	at org.broad.igv.ui.panel.ZoomSliderPanel.paintComponent(ZoomSliderPanel.java:122)
```

lbergelson:
@jrobinso Good catch.  I'll look into that.

lbergelson:
@jrobinso I believe I have fixed the issue.  It seemed to occur when the feature window was larger than the current chromosome length causing it to produce a negative log value.  I capped the window length when calculating the zoom now so there shouldn't be anymore negative numbers there.

jrobinso:
Looks good now

PR #1200, opened 2022-08-24
Base mod likelihood comp bug

jrobinso:


jrobinso:
Fix base modification likelihood comparison bug (see #1200)

PR #1245, opened 2022-10-25
Update AlignmentTrack.java

CJ-Chen:
"insertionRect" needs to be initialized though not displayed.
It might cause a null pointer exception in Line 709 "Rectangle iRect = new Rectangle(inputRect.x, insertionRect.y, inputRect.width, insertionRect.height);"
when some alignment tracks are displaying insertion markers while others do not.

### DIFF
--- a/src/main/java/org/broad/igv/sashimi/SashimiContentPane.java
+++ b/src/main/java/org/broad/igv/sashimi/SashimiContentPane.java
@@ -5,7 +5,11 @@ import org.broad.igv.ui.panel.Paintable;
 import javax.swing.*;
 import java.awt.*;
 
-public class SashimiContentPane extends JPanel implements Paintable {
+public class SashimiContentPane extends JSplitPane implements Paintable {
+
+    public SashimiContentPane(JPanel sashimiPanel, JScrollPane scrollableGenePane) {
+        super(JSplitPane.VERTICAL_SPLIT, sashimiPanel, scrollableGenePane);
+    }
 
     @Override
     public void paintOffscreen(Graphics2D g, Rectangle rect, boolean batch) {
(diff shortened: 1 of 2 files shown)


CJ-Chen:
Sorry for this PR.
It seems that this modification will consume much more time for graph rendering. 
I don't know why.

CJ-Chen:
Maybe we can use "insertionRect.setBounds(rect);" to replace "insertionRect = new Rectangle(rect);" around line 523?

jrobinso:
I'll investigate when I can find time, but do you have steps to reproduce an issue?  In what situation did an error occur?

CJ-Chen:
> I'll investigate when I can find time, but do you have steps to reproduce an issue? In what situation did an error occur?

Here is a genome with two alignment track. One is displaying with “Show Insertion Mark” and the other without selected this option.



And when i click this the insertion mark. It will report a null pointer exception that mention above.



```
igv.log:
SEVERE [10月 27,2022 14:00] [DataPanel] Error painting DataPanel
SEVERE [10月 27,2022 14:00] [DataPanel] java.lang.NullPointerException
	at org.igv/org.broad.igv.sam.AlignmentTrack.renderExpandedInsertion(AlignmentTrack.java:679)
	at org.igv/org.broad.igv.ui.panel.DataPanelPainter.paintExpandedInsertion(DataPanelPainter.java:214)
	at org.igv/org.broad.igv.ui.panel.DataPanelPainter.paint(DataPanelPainter.java:89)
	at org.igv/org.broad.igv.ui.panel.DataPanel.paintComponent(DataPanel.java:154)
```

Sorry for the inconvenience.

jrobinso:
What version of IGV are you using?  I can't reproduce this error, and the line numbers in the igv.log stack trace do not match the current code.

jrobinso:
I can't reproduce the error just loading tracks, only after loading tracks via a session.  So I think this is the same error as reported here: https://github.com/igvteam/igv/issues/1252.   I will close this PR as a duplicate bug report.

PR #1248, opened 2022-10-26
Refactoring BisulfiteContext enum to reduce code duplication

lbergelson:
I moved a bunch of maps and functions that were used with the enum into actually being part of the enum.  This should remove clutter and make it clearer.


jrobinso:
Looks much better.  I pushed a test session for bisfulfite data, comparisons with release 2.15.1 look identical.

Issue #738, opened 2019-12-03
Split screen by supplementary alignments


jrobinso:
See #298

ryan-moreno:
I am trying to access this feature and am having difficulty. Based on #1278, my understanding is that if I right-click on an aligned read that has a supplementary alignment tag, then I should get the option to split screen. However, this is not among the options I'm seeing. I have used various preferences and tried to find documentation for this feature, but have not found the solution. My use case is that I want to view the two regions surrounding the ends of a translocation and see which reads span the translocation. I can pull up the two regions in split screen manually, but using "Link supplementary alignments" doesn't seem to work across split screen. Any suggestions?

jrobinso:
What version of IGV are you running?

jrobinso:
And linking (as in drawing a connecting line) will not work across split screens, they are completely independent.

ryan-moreno:
I am using 2.16.0. Happy to provide log files if that is helpful.

jrobinso:
I'm not entirely sure what you are trying to do, but there are a number of features for split reads in the latest version  (2.18).  See https://igv.org/doc/desktop/#UserGuide/tracks/alignments/chimeric_reads/.    I do not recall exactly what was in 2.16 in this regard.

PR #1249, opened 2022-10-26
Fixing the treatment of PairedAlignment so gotoMate works.

lbergelson:
I noticed that 'gotoMate' didn't work when in linked pairs mode, but 'view mate region' in split screen did.  This generalizes the logic so that the same conditions apply to both.  It helps prepare for https://github.com/igvteam/igv/issues/738

* Adding a default method getSpecificAlignment(location) to Alignment which returns the most specific sub alignment at a location, it is specialized in PairedAlignment and LinkedAlignment
* Cleaning up and unifying code so that gotoMate and viewMateInSplit screen work in the same circumstances
* Adding overloads and changing parameters in several methods to reduce boilerplate.

### DIFF
--- a/src/main/java/org/broad/igv/sam/AlignmentTrack.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentTrack.java
@@ -853,6 +853,13 @@ public class AlignmentTrack extends AbstractTrack implements IGVEventObserver {
     }
 
 
+
+    Alignment getAlignmentAt(final TrackClickEvent te) {
+        MouseEvent e = te.getMouseEvent();
+        final ReferenceFrame frame = te.getFrame();
+        return frame == null ? null : getAlignmentAt(frame.getChromosomePosition(e), e.getY(), frame);
+    }
+
     Alignment getAlignmentAt(double position, int y, ReferenceFrame frame) {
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/broad/igv/sam/AlignmentTrack.java
```diff
@@ -885,12 +892,16 @@ public boolean handleDataClick(TrackClickEvent te) {
...
-                IGV.getInstance().repaint(this);
-                return true;
+            Alignment alignment = this.getAlignmentAt(te);
+            if (alignment != null) {
+                if (selectedReadNames.containsKey(alignment.getReadName())) {
+                    selectedReadNames.remove(alignment.getReadName());
+                } else {
+                    setSelectedAlignment(alignment);
+                }
+                IGV.getInstance().repaint(this); //todo check if doing this conditionally here is ok
```
lbergelson, line 902: The logic here was inlined from selectAlignment(e, frame).  The repaint logic changed slightly, previously it would repaint when no alignment was present at the given location but now it only does so if an alignment was selected.  I assume that's ok but i figured I would ask to be sure I didn't miss something.

jrobinso: That looks o.k.

PR #1260, opened 2022-11-18
Fixes a URL query parsing bug for Vcf to BAM mapping

ubuntolog:
I have found another bug in URL parsing. My authentication token passed inside a URL was treated as a part of the filename and I was getting Error 403. I have implemented a simple fix.


jrobinso:
Thanks for the PR.  I wasn't aware anyone used the Vcf to BAM mapping, this was implemented for a system that was deprecated years ago.

ubuntolog:
Sure, no problem. I noticed a lot of error messages in my server logs when IGV was checking if such files existed.

PR #1267, opened 2022-12-12
Fixing issue where non-snp Variants were not properly selectable

lbergelson:
* There was a bug causing multibase variants to only be selectable when clicking on the first base.
  They should now be clickable at any position
  This was particularly noticeable in gvcfs where there are many long reference blocks.


jrobinso:
Looks good, thanks for the unit tests.

PR #1284, opened 2023-02-14
Adding Access-Control-Allow-Headers

tomkp75:
Adding Access-Control-Allow-Headers to sendHTTPOptionsResponse for options preflight to pass when opening IGV through an HTML link using a HTTP Client


jrobinso:
Interesting, what is triggering preflight requests?

jrobinso:
Also, I think you need to explicitly print CRLF,  not use println,  for this to work on all platforms.

tomkp75:
Thanks @jrobinso I updated the code. I'm calling IGV using axios (rf. https://axios-http.com/docs/intro) which triggers a preflight request. 
I could use window.open but it's not as clean and bind IGV to the current window (or new tab) which is not necessary.

jrobinso:
OK merged.

tomkp75:
Thanks @jrobinso. Will update IGV on the next release. When should we expect it?

jrobinso:
Probably mid March.  It should be available in the development snapshot build now

PR #1309, opened 2023-04-03
Session autosaving on shutdown and optional autoloading on startup

### LINKED ISSUE #1296: Feature request - highlighting location of matching soft-clipped reads
Dear IGV team, 

I use IGV daily in my research in rare disease and find it a fantastic tool - so thank you !

The feature I would like to see is a way of highlighting the location of a series of soft-clipped reads, located at the same location in the genome, with at least 6-10bp of matching soft-clipped sequence,   so that you can see their location in the genome without having to zoom in to the base pair level.

Ideally the feature would allow you to select how many soft-clipped reads start in the same location to be marked, as well as how many reads in the soft-clipped sequence need to match each other, again to be marked.   This would be highlighting breakpoints in the DNA (or RNA), that have a certain amount of reads supporting it. How these breakpoints are highlighted would need to be different than the usual purple arrow for small mapped insertions ( ie dup of a nucleotide) 

The reason I would like this feature is one of the more difficult mutations to pick up are insertions. They are also very hard to spot in IGV unless you zoom right in to the base pair level.    

I have pasted an example which I would like IGV to be able to highlighted. 
 - notice that the soft clipping all starts at the same location 
 - and the sequence of the soft-clipping is the same, though there are different lengths of soft-clipping in each read.  
image. 
- it is also worth noting that many of the examples I would like highlighted, have inserted sequence that do not map elsewhere in the genome ( or not uniquely). For example, novel insertions of ALU  sequences.

Is this a feature you think you might be able to include in IGV?  I acknowledge that confirming the soft-clipped sequences match each other is not an easy / quick problem to solve.  

Please let me know if you would like more information or examples

Kind regards,
Katrina

KevinCLydon:
Autosaves session to a file in the igv folder when exiting.  Added option in preferences to enable automatically loading autosave when starting IGV.

Modified quit handler to not call shutdown twice, which seemed to cause the saving to happen twice.

I didn't see any unit tests associated with the files I changed, so does it make sense to add unit tests?

Closes #1296


jrobinso:
This is not the sort of thing you can unit test easily.

PR #1315, opened 2023-04-10
Update to use Java 17

lbergelson:
This updates to use / require java 17.  
I first tried to update the test module settings and ran into trouble.  It seems like removing all of the test module settings completely works though.  I set it to use the gradle junit platform settings which may be doing that in the background now.  

Tests seem to run locally and on github without problems so I think it seems to be working.

I also update the versions of the github actions since they were using deprecated versions.


build.gradle
```diff
@@ -127,10 +133,13 @@ dependencies {
     )
 
     testImplementation(
-            [group: 'junit', name: 'junit', version: '4.12'],
+            [group: 'junit', name: 'junit', version: '4.13'],
             [group: 'com.sparkjava', name: 'spark-core', version: '2.2'],
             [group: 'org.glassfish.jersey.core', name: 'jersey-common', version: '2.22.4']
     )
+    testRuntimeOnly(
+           [group: 'org.junit.vintage', name:'junit-vintage-engine', version:'5.8.2']
```
lbergelson, line 135: This test runtime dependency is the engine that runs junit4 tests since gradle seems to assume you're using 5 now.

build.gradle
```diff
@@ -243,32 +224,32 @@ tasks.distTar.enabled = false
 tasks.startScripts.enabled = false
 
 task createDistZip(type: Zip, dependsOn: createDist) {
-    archiveName = "IGV_${version}.zip"
+    archiveFileName = "IGV_${archiveVersion}.zip"
```
lbergelson, line 222: The name changes are all to satisfy gradle deprecation complaints.

src/main/java/org/broad/igv/prefs/IGVPreferences.java
```diff
@@ -248,9 +249,9 @@ public boolean getAntiAliasing() {
     private void updateCaches(String key, String value) {
         key = key.trim();
         if (booleanCache.containsKey(key)) {
-            booleanCache.put(key, new Boolean(value));
+            booleanCache.put(key, Boolean.valueOf(value));
         }
-        colorScaleCache.remove(key);
+        colorScaleCache.remove(key); //TODO color scale cache doesn't use key strings so something here is wrong.
```
lbergelson, line 258: Intellij pointed out that these keys are the wrong type so it seems very likely that this cache isn't behaving as expected.  We should probably address that seperately.

davideby:
Looks good to me.  Glad to see those hacks in `compileTestJava` are no longer necessary.

I'll note there a warning / non-fatal error when building on my Linux laptop:
```
Invalid Java installation found at '/usr/lib/jvm/openjdk-11' (Common Linux Locations). It will be re-checked in the next build. This might have performance impact if it keeps failing. Run the 'javaToolchains' task for more details.
Invalid Java installation found at '/usr/lib/jvm/openjdk-17' (Common Linux Locations). It will be re-checked in the next build. This might have performance impact if it keeps failing. Run the 'javaToolchains' task for more details.
Invalid Java installation found at '/usr/lib/jvm/openjdk-16' (Common Linux Locations). It will be re-checked in the next build. This might have performance impact if it keeps failing. Run the 'javaToolchains' task for more details.
```

I have no idea if Mac or Windows builds report the same.

This seems to be a [Gradle bug](https://github.com/gradle/gradle/issues/22921) that they claim is Closed.  My take is that this doesn't affect anything and can be ignored at least for now, but you might want to keep an eye on it.

hubutui:
Great job! But I encounter this error when I run `gradle createDist` on Linux:
```text
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':createDist'.
> Resolving dependency configuration 'default' is not allowed as it is defined as 'canBeResolved=false'.
  Instead, a resolvable ('canBeResolved=true') dependency configuration that extends 'default' should be resolved.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.3/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
```
`./gradlew createDist` gave me a similar error:
```text
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':createDist'.
> Resolving dependency configuration 'default' is not allowed as it is defined as 'canBeResolved=false'.
  Instead, a resolvable ('canBeResolved=true') dependency configuration that extends 'default' should be resolved.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

jrobinso:
@hubutui the "createDist" problem is fixed.

build.gradle
jrobinso: OK, looks good

src/main/java/org/broad/igv/prefs/IGVPreferences.java
jrobinso: OK I opened an issue and assigned to myself

jrobinso (approved):

PR #1327, opened 2023-05-12
Timed autosaves of sessions

KevinCLydon:
The use cases I'm trying to address here are:

1. IGV crashes or otherwise closes in a non-standard way so it does not autosave on exit, and the user wants to recover their session (or at least a not-too-old version of it).
2. A user realizes they want to return to a version of their session from a few minutes ago.  This one is maybe covered more as a side effect of covering the first use case.

To these ends, I made some changes to how autosaving works:

1. Moved autosave files to an autosave folder within the igv directory.  
3. Now there can be multiple and they are named with a datetime for when they were created (e.g. `session_autosave2023-05-05T15:54:08.681Z.xml`).
4. Autosaves will be created every x minutes (along with the autosave that will be created when you exit), where x is a configurable preference called `AUTOSAVE_FREQUENCY`.  It will retain y autosaves at a time, where y is a configurable preference called `AUTOSAVES_TO_KEEP`.  As part of the autosave process, when a new autosave is created, IGV will delete the oldest autosave if the number of autosave files exceeds y.
    1. Both of these preferences can be set in the preferences menu under a "Session Autosave" subhead.
    3. AUTOSAVE_FREQUENCY defaults to 10 minutes.
    4. AUTOSAVES_TO_KEEP defaults to 5.
5. There is now a check when loading the recent sessions list at startup that prevents sessions from being added if the file no longer exists.

I do have some questions before I'll feel like it's ready for an actual review and merge:

1. Right now, it is possible that the user could try to load an autosave while it is being deleted.  If I'm correct, this will result in an error message being displayed to the user.  Is this sufficient, or should I add a mutex or some similar solution to prevent the autosave mechanism from deleting files while the user is attempting to load a session?
2. Are the AUTOSAVE_FREQUENCY and AUTOSAVES_TO_KEEP defaults sensible?
3. Should it be possible to disable the timed autosave?
5. I'm not sure how to handle the recent sessions list with this setup.  The last autosave created on exit will appear in the recent sessions list when a user opens IGV, but it does not disappear from the list if it's deleted as part of the timed autosave process, which could cause the user to try to load it unsuccessfully.
    1. Also, currently, I'm only putting the autosave that's created on exit in that list.  I'm not sure if it makes more sense to just include the most recent autosave maybe?  Ideally, I'd like the user to be able to see all the autosaves that still exist, but I don't know if the recent sessions list is the place for that, since it would quickly become cluttered.


lbergelson:
Is the save on exit autosave included in the autosaves to keep list / deleted over time?  I might keep the autosaves generated from the save on exit separate from the timed ones. I.e. just keep 1 that's only deleted on exit and separate from the rolling ones.

Don't think it make sense to add the timed autosaves to the recent list.  I would maybe add a single "autosaves" option which expands into a live list of the currently available ones.

KevinCLydon:
Right now the exit autosave is kept in the same list as the timed autosaves.  That's a good idea, though.  I'll split them up.  and the autosaves option for seeing the other ones makes sense.  That should be easy enough to do.  Thanks!

jrobinso:
RE 3,  yes, and not only should it be possible, but maybe should be off by default.  In any event this can be controlled with AUTOSAVE_FREQUENCY without introducing another boolean flag.

jrobinso:
If I load an autosave through the menu,  then select "Save Session", it defaults to the same filename.    That's maybe o.k.,  but I think no assumption should be made when saving.  In other words the default "igv_session.xml" name (or whatever the default is).

Also, the forward slashes in the file names could cause issues on some platforms (forward slasshes are in the dates).   These should maybe be changed to underscores.

KevinCLydon:
@jrobinso I think it should be all ready now

jrobinso:
@KevinCLydon Looks good!

PR #1331, opened 2023-05-15
Adding improved support for displaying chimeric / supplementary reads

lbergelson:
* When reads include an SA tag it will now be used to add additional information Color clipped ends by the associated supplementary reads location
	* within a chromosome color by inversion/non-inversion * connections between chromosomes are colored by the associated chromosome * new preference to display the nane the connected chromosome
* new grouping option to group reads that have an SA tag vs not
* new sort options to sort reads by the amount of clipping at either end
* New diagram popup to visually show the connections between a single reads set of supplementary reads
* Improved supplementary read information in the tooltip
* Splitting by SATag / Mates now displays the selected reads at the top of the window * fixed some issues around frame loading to support this where each frame would force a full repack of all the other frames when it finished loading

* Other changes and additions:
  * Added a new ClippingCounts class to standarize how clipping information is counted and passed around
  * Many additions to SupplementaryAlignment
  * Changes to Alignment and SAMAlignment to support more efficient CIGAR access and caching of expensive  computed values
  * Select read by name now prefils the name of the clicked read if there is one
  * Fix frame sorting issue
  * Improve frame loading so refreshing a single frame doesn't force a repack of every frame.
  * Reactived event that fires on data load which is used to complete sorts that
  * Triggering a sort when all frames are not fully loaded will now queue the sort and perform it when the frame is done loading
  * make STOP button work to abort loading reads, it may end up in a weird state, but it's the same state you get when it stops itself due to memory issues, this just lets you get there without waiting for memory to fill up
  * delete unused RenderReset class


lbergelson:
@jrobinso Here it is, much overdue.  It does a bunch of useful stuff but there are many more things that could be added.  Take a look and let me know if there are things you want to change.  

I hope to push some of the ClippingCounts stuff down to htsjdk because we compute that over and over in all sorts of places in our various codebases and I think it could be generally useful.  

There are lots of things I want the new popup window to do / allow for interaction.  Ideally it probably shouldn't be it's own unique thing but should use the existing AlignmentRenderer / possibly share code with the SashimiPlot stuff.

build.gradle
```diff
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2007-2019 Broad Institute
+// Copyright (c) 2007-2019 Broad Institut≈e
```
droazen, line 3: Looks like a stray character got introduced here... :)

jrobinso:
@lbergelson I'm not sure "sorting selected chimeric alignments" or whataever we call that is quite working yet.   In the pic below, the selected reads on the right were at the top while the frame on the left was being loaded,   but after loading they got shoved down a bit.    Also, as an aide, there appears to be duplicates in this file.

Screen Shot 2023-05-18 at 11 26 14 AM

PR #1339, opened 2023-06-06
Abbreviate CIGAR strings in popups

lbergelson:
As we discussed a while ago, this limits the length of the displayed CIGAR string in text popups.

This capability already existed it turned out, but the limit was set to 1000 which seems way to long.  I set it to 50 instead which looks good to me, but I'm happy to tweak it.

I also moved some regex patterns into constants so we don't have to pay to recompile them repeatedly.


lbergelson:
It looks like this now.
Screenshot 2023-06-06 at 5 24 54 PM

PR #1342, opened 2023-06-13
Add a dedicated 6mA color mode

ctsa:
This allows a user to visualize 6mA modifications only when viewing a bam file containing multiple modification types.

----

This is an initial proposal for a dedicated 6mA visualization option. This has come up in the context of Fiber-Seq samples, where the bam is annotated with both 5mC and 6mA modifications, and it can be useful to view the 6mA modifications only.

### DIFF
--- a/src/main/java/org/broad/igv/sam/AlignmentRenderer.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentRenderer.java
@@ -1301,6 +1301,7 @@ public class AlignmentRenderer {
             case BASE_MODIFICATION:
             case BASE_MODIFICATION_5MC:
             case BASE_MODIFICATION_C:
+            case BASE_MODIFICATION_6MA:
             case SMRT_SUBREAD_IPD:
             case SMRT_SUBREAD_PW:
             case SMRT_CCS_FWD_IPD:
--- a/src/main/java/org/broad/igv/sam/AlignmentTrack.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentTrack.java
@@ -103,6 +103,7 @@ public class AlignmentTrack extends AbstractTrack implements IGVEventObserver {
         BASE_MODIFICATION,
         BASE_MODIFICATION_5MC,
         BASE_MODIFICATION_C,
+        BASE_MODIFICATION_6MA,
         SMRT_SUBREAD_IPD,
         SMRT_SUBREAD_PW,
         SMRT_CCS_FWD_IPD,
@@ -111,7 +112,7 @@ public class AlignmentTrack extends AbstractTrack implements IGVEventObserver {
         SMRT_CCS_REV_PW;
 
         public boolean isBaseMod() {
-            return this == BASE_MODIFICATION || this == BASE_MODIFICATION_5MC || this == BASE_MODIFICATION_C;
+            return this == BASE_MODIFICATION || this == BASE_MODIFICATION_5MC || this == BASE_MODIFICATION_C || this == BASE_MODIFICATION_6MA;
         }
 
         public boolean isSMRTKinetics() {
--- a/src/main/java/org/broad/igv/sam/AlignmentTrackMenu.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentTrackMenu.java
@@ -636,6 +636,7 @@ class AlignmentTrackMenu extends IGVPopupMenu {
         mappings.put("base modification", AlignmentTrack.ColorOption.BASE_MODIFICATION);
         mappings.put("base modification (5mC)", AlignmentTrack.ColorOption.BASE_MODIFICATION_5MC);
         mappings.put("base modification (all C)", AlignmentTrack.ColorOption.BASE_MODIFICATION_C);
+        mappings.put("base modification (6mA)", AlignmentTrack.ColorOption.BASE_MODIFICATION_6MA);
         colorMenu.addSeparator();
         for (Map.Entry<String, AlignmentTrack.ColorOption> el : mappings.entrySet()) {
             JRadioButtonMenuItem mi = getColorMenuItem(el.getKey(), el.getValue());
--- a/src/main/java/org/broad/igv/sam/mods/BaseModificationCoverageRenderer.java
+++ b/src/main/java/org/broad/igv/sam/mods/BaseModificationCoverageRenderer.java
@@ -32,10 +32,10 @@ public class BaseModificationCoverageRenderer {
                 draw5MC(context, pX, pBottom, dX, barHeight, pos, alignmentCounts, true);
                 break;
             case BASE_MODIFICATION_6MA:
-                draw6MA(context, pX, pBottom, dX, barHeight, pos, alignmentCounts);
+                draw(context, pX, pBottom, dX, barHeight, pos, alignmentCounts, true);
                 break;
             default:
-                draw(context, pX, pBottom, dX, barHeight, pos, alignmentCounts);
+                draw(context, pX, pBottom, dX, barHeight, pos, alignmentCounts, false);
         }
     }
 
@@ -46,7 +46,8 @@ public class BaseModificationCoverageRenderer {
                              int dX,
                              int barHeight,
                              int pos,
-                             AlignmentCounts alignmentCounts) {
+                             AlignmentCounts alignmentCounts,
+                             boolean onlyDraw6mA) {
 
         BaseModificationCounts modificationCounts = alignmentCounts.getModifiedBaseCounts();
 
@@ -56,6 +57,12 @@ public class BaseModificationCoverageRenderer {
 
             for (BaseModificationCounts.Key key : modificationCounts.getAllModifications()) {
 
+                String modification = key.getModification();
+                if (onlyDraw6mA) {
+                    if (key.getCanonicalBase() != 'A' && key.getCanonicalBase() != 'T') continue;
+                    if (!modification.equals("a")) continue;
+                }
+
                 // The number of modification calls, some of which might have likelihood of zero
                 int modificationCount = modificationCounts.getCount(pos, key);
 
@@ -63,8 +70,6 @@ public class BaseModificationCoverageRenderer {
 
                     byte base = (byte) key.getBase();
                     byte complement = SequenceUtil.complement(base);
-                    char modStrand = key.getStrand();
-                    String modification = key.getModification();
 
                     // Count of bases at this location that could potentially be modified, accounting for strand
                     int baseCount = alignmentCounts.getPosCount(pos, base) + alignmentCounts.getNegCount(pos, complement);
@@ -155,82 +160,6 @@ public class BaseModificationCoverageRenderer {
                 Arrays.sort(orderedMods, (o1, o2) -> -1 * o1.compareTo(o2));
                 for (String m : orderedMods) {
                     Color mColor = BaseModificationColors.getModColor(m, (byte) 255, ColorOption.BASE_MODIFICATION_5MC);
-                    int mModHeight = (int) Math.round(((likelihoodSums.get(m)) / t) * calledBarHeight);
-                    if (mModHeight > 0) {
-                        baseY -= mModHeight;
-                        graphics.setColor(mColor);
-                        graphics.fillRect(pX, baseY, dX, mModHeight);
-                    }
-                }
-            }
-        }
-    }
-
-    private static void draw6MA(RenderContext context,
-                                int pX,
-                                int pBottom,
-                                int dX,
-                                int barHeight,
-                                int pos,
-                                AlignmentCounts alignmentCounts) {
-
-        BaseModificationCounts modificationCounts = alignmentCounts.getModifiedBaseCounts();
-
-        if (modificationCounts != null) {
-
-            final byte base = (byte) 'A';
-            final byte complement = (byte) 'T';
-
-            Map<String, Integer> likelihoodSums = new HashMap<>();
-            Map<String, Integer> modCounts = new HashMap<>();
-            for (BaseModificationCounts.Key key : modificationCounts.getAllModifications()) {
-
-                if (key.getCanonicalBase() != 'A' && key.getCanonicalBase() != 'T') continue;
-                if (key.getModification().equals("a")) {
-                    String mod = key.getModification();
-                    final int count = modificationCounts.getCount(pos, key);
-                    if (count > 0) {
-                        modCounts.put(mod, count);
-                        final int likelhoodSum = modificationCounts.getLikelhoodSum(pos, key);
-                        likelihoodSums.put(mod, likelhoodSum);
-                    }
-                }
-            }
-
-            if (likelihoodSums.size() > 0) {
-
-                // Count of bases at this location that could potentially be modified
-                double modifiableBaseCount = alignmentCounts.getPosCount(pos, base) + alignmentCounts.getNegCount(pos, complement);
-
-                // Compute "snp factor", ratio of count of base calls that could be modfied (on either strand) to
-                // total count. This is normally close to 1, but can be less due non CG bases at this location (e.g. snps)
-                double cgCount = alignmentCounts.getCount(pos, base) + alignmentCounts.getCount(pos, complement);
-                double snpFactor = cgCount / alignmentCounts.getTotalCount(pos);
-
-                double calledBarHeight = snpFactor * barHeight;
-                double t = modifiableBaseCount * 255;   // If all bases are called this is the total sum of all likelihoods, including "no mod" likelihood
-
-                // Likelihood of no modification.  It is assumed that the likelihood of no modification == (1 - sum(likelihood))
-                int c = Collections.max(modCounts.values());
-                double noModProb = c * 255;
-                for (String m : modCounts.keySet()) {
-                    if (likelihoodSums.containsKey(m)) {
-                        noModProb -= likelihoodSums.get(m);
-                    }
-                }
-
-                // Draw "no mod" bar
-                int noModHeight = (int) Math.round((noModProb / t) * calledBarHeight);
-                int baseY = pBottom - noModHeight;
-                Graphics2D graphics = context.getGraphics();
-                graphics.setColor(BaseModificationColors.noModColor5MC);
-                graphics.fillRect(pX, baseY, dX, noModHeight);
-
-                // Loop through modifications drawing bar for each
-                String[] orderedMods = likelihoodSums.keySet().toArray(new String[0]);
-                Arrays.sort(orderedMods, (o1, o2) -> -1 * o1.compareTo(o2));
-                for (String m : orderedMods) {
-                    Color mColor = BaseModificationColors.getModColor(m, (byte) 255, ColorOption.BASE_MODIFICATION_6MA);
                     int mModHeight = (int) Math.round(((likelihoodSums.get(m)) / t) * calledBarHeight);
                     if (mModHeight > 0) {
                         baseY -= mModHeight;
--- a/src/main/java/org/broad/igv/sam/mods/BaseModificationCoverageRenderer.java
+++ b/src/main/java/org/broad/igv/sam/mods/BaseModificationCoverageRenderer.java
@@ -31,6 +31,9 @@ public class BaseModificationCoverageRenderer {
             case BASE_MODIFICATION_C:
                 draw5MC(context, pX, pBottom, dX, barHeight, pos, alignmentCounts, true);
                 break;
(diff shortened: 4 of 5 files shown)


jrobinso:
Thanks @ctsa .  This is timely, I am porting this to igv.js right now.    

What % of this code is a copy of the 5mC drawing functions?   Could it be merged with some flag on what modification you want to see?    I have some concern about code bloat and scaling this.   

Do you have some test data I can access?

ctsa:
Hi Jim,

It is a big copy/paste right now just to test out the idea (and we had an immediate use case). I could take a pass at some consolidation if you prefer?

For test data, I think a small test bamlet should be something we can produce in the short term -- we might be able to work with @mrvollger for this?

jrobinso:
OK, got it.     Before merging I think we should consolidate if it makes sense.

As you probably know, the 5mC mode is implemented to mimic to an extant the already existing bisulfite coloring scheme.  Likelihood < 50% is interpreted as evidence the modification is absent, and gets a distinct color (blue).    In the original experiments motivating the bisulfite scheme the lack of methylation (e.g. loss in a tumor sample) was as significant or often more significant than its presence.    If this view is good for other experiments we can generalize it, as you're doing.  The alternative approach is to use a single color with alpha.   They convey the same information but one emphasizes lack of evidence as well as its presence equally.   Does that make sense?

jrobinso:
That is, as opposed to generalizing somehow the special 5mC mode.

ctsa:
Thanks @jrobinso - The scope I was limiting this branch to was to provide the same 6mA visualization you would get if there were not 5mC (or other interfering basemods) in the bam. Right now this is "one color + alpha".

In this way the currently proposed 6mA viewing mode is not analogous to the special 2-color 5mC mode. It is more like a filter, letting you focus exclusively on the 6mA patterns in the bam without other basemods visually interfering.

jrobinso:
@ctsa got it.   So I think what we need is perhaps an extra argument to the generic draw function, i.e. this one

'''
draw(context, pX, pBottom, dX, barHeight, pos, alignmentCounts);
'''

which specifies a specific modification, defaulting to "all",  and some extra stuff in the menu to let you select it.   If a specific modification is specified others are ignored.   I think all we need to do is modify this,  around line 162 of BaseModificationRenderer.   If there are multiple modifications this selects the most likely.   Instead we would select the one specified (e.g. 6mA).   There's probably a more direct way to do it but I think we just need an argument to "draw" to specify the modification of interest.  If the argument is null it could do what it does now.   The coverage renderer could be modified similarly, I think there we just don't draw the slices for other modifications.   Does this sound reasonable?

```
                                if (modification == null || l > lh) {
                                    modification = bmSet.getModification();
                                    lh = l;
                                }
```

jrobinso:
When I get a test file with 6mA I'll look at this, again this is timely as I'm in this code porting it to igv.js.

mrvollger:
@jrobinso here is a small bam file against hg38, but note it only has data for this region:
chr11:119,076,212-119,102,218 
Let me know if it works for you. 
[Archive.zip](https://github.com/igvteam/igv/files/11749590/Archive.zip)

jrobinso:
@mrvollger Yes that works.  I observe the color used for 6ma is rather weak (grayish green), maybe we should address that too.

I will get to this in the next day or two.

mrvollger:
Awesome! I very much agree on the color not being great and if you are open to suggestions we like using purple (specifically 51,0,111) for m6a marks in our lab. 

Thanks!

ctsa:
@jrobinso - I've refactored to compress down the code duplication, and you've suggested. Now the standard draw functions handle 6mA mode as a special case.

Testing this on Mitchell's example bamlet I see the expected behavior, with standard base modification mode showing the 5mC/6mA combination:

image

...and the new 6mA mode showing just 6mA:

image

jrobinso:
I'm going to merge this.   I noticed an issue with the coverage track, but I don't think its new.  The height of the coverage bar for modifications can be > 100% of the total height, so to speak.

There's also something I don't understand about this data.   There seems to be 6ma modifications on canonical T bases on the + strand, and conversely on canonical A bases on the - strand.  I wouldn't expect either of these.   Is this expected, or should we look deeper for an IGV bug here? 

Screen Shot 2023-06-17 at 2 57 54 PM

jrobinso:
Actually these might be related.

mrvollger:
There should for sure only be A+ and T- and if that’s not the case in the read coordinates I have a bug to look into… I am away at the moment but do you think this might be a liftover issue between the reads and the reference that makes it look this way? Or do you think it’s an error in the raw data?

mrvollger:
Checked the raw data and it looks fine, so possibly an IGC bug:
```
import pysam
import sys

bam = pysam.AlignmentFile("HG002_chr11_119076212_119102218.bam")

for rec in bam.fetch(until_eof=True):
    seq = rec.get_forward_sequence()
    mods = rec.modified_bases_forward
    for (base, strand, mod), pos_qual_list in mods.items():
        if mod != "a":
            continue
        for pos, qual in pos_qual_list:
            if strand == "+":
                assert seq[pos] == "A", seq[pos]
            elif strand == "-":
                assert seq[pos] == "T", seq[pos]
            else:
                sys.stderr.write("\rfine")

```

jrobinso:
There is something wrong.  This is the first dataset I've looked at with both "A+" and "T-" calls.   For my own education, how are these different?    I'm mostly looked at 5mC datsets,  encoded as "C+m",  of course you get calls on + strand alignments at "C" positions and at "G" positions on - strand alignments.     I've not yet seen a dataset with "C+" and "G-" calls, but I only look at these when there is a problem.

I will look into this, which means understanding that SPEC again arggh.

jrobinso:
The error hear has nothing to do with A+a and T-a calls, although that does create 2 sets of modifications in IGV that are logically equivalent for visualization it shouldn't case an issue.   The problem I think is the treatment of the "-" call, although there is a unit test for that.    I've never felt comfortable with that spec, I have no idea what "top" and "bottom" strand means and have probably misinterpreted this.   Very surprising there are no bug reports on this.

mrvollger:
I just wrote the code that was encoding these m6a calls so I needed to make sure there wasn't an error in my encoding. 

So PacBio CCS/HiFi sequences both strands of a DNA molecule; however, they are computationally combined into one output in the bam format. This means that it can give you information about an m6A mod that happens on the opposite strand as the one reported in the sequence field in the bam format. And this is the case where you should use T-a.

The difference from CpG here is the rev comp of CpG is CpG and that CpG methylation is mostly always seen on both strands at once to my understanding. 

Does this help?

jrobinso:
I think I confused myself, and there is not (necessarily) a problem at all.   All the unit tests, which came fro James Bonfield's set, pass.   Its o.k., of course, to have a 6mA call on a "T" if its called on the other strand.    So never mind, unless you find an issue.   Look very carefully though.    Maybe if we both scrutinize the same alignment we can confirm we are o.k.,  I'm lookin at `m84039_230418_220405_s4/237177301/ccs`.   

I still have never got my head around "top" and "bottom" terminology in the samtags spec.

I do think my math for the coverage track might be wrong,  but the alignments are (probably) o.k.

jrobinso:
BTW if you want to confirm your encodings the test files (and associated .txt) here are really valuable,  I would never have understood that spec without them. If indeed I ever have understood that spec.

https://github.com/samtools/hts-specs/tree/master/test/SAMtags

jrobinso:
The conventions for CpG might carry over from bisulfite sequencing, where there is never the equivalent of a "G-" call,  so only ~1/2 the reads at any position have any possibility of being called (assuming equal strand coverage).    The early base modification files for 5mC seemed to be like this too,  and there were complaints that the coverage only showed ~50% called (again assuming equal strand coverage) when in fact it should be 100%, considering that 1/2 of the reads were not informative.   The ability to sequence both strands is a new wrinkle.   But this should only affect the coverage track,  I need to look carefully at that.   The alignments should (might?) be o.k.

ctsa:
Hi @jrobinso, Thanks for merging!

Just catching up on today's thread. The 6ma visualization I've seen so far look good.

Regarding your first comment today with the IGV image, what might be complicating things here are two different uses of the word strand. The base modifications refer to the + and - strands, meaning 'the strand reported by the sequencer' (+) and its complement (-). This should be distinguished form the reference alignment strand.

Regarding the confusing representation of 5mC at CpG sites, they would be more correctly reported as paired C+/G- calls in the bam MM tag -- but so far the commonly accepted encoding in the MM tag has been to shortcut this step and enter one C+ tag per CpG. One reason why they haven't been entered correctly as C+/G- pairs is that that would provide no way to express that the confidence of the base modification in the ML tag was a joint confidence over the C+/G- pair.

jrobinso:
Thanks @ctsa and @mrvollger ,   I think everything is actually o.k.  wrt the alignments.  I do need to rethink the coverage track and what we want to convey there.   It was heavily influenced by 5MC/CpG conventions.

jrobinso:
Well slight caveat,  the positions of modifications on the alignments should be correct.   The likelihoods now are complicated if these is both a T- and A+ report at the same position.   What the code is doing now is, in effect, using the maximum likelihood.   Is that the right thing?  I doubt it but what should we do?   I think @ctsa alluded to this wrt 5mC, its taken care of in that case by combing the C+/G-  and reporting a single likelihood

mrvollger:
So if my understanding is correct this shouldn't be possible and T- and A+ should be mutually exclusive for m6A. I think the only reason CpG has this issue is that the reverse complement of it is also CpG. 

If that's not the case let me know Because I think it should not be possible.

jrobinso:
@mrvollger OK,  I was just thinking through the consequences of this,  "So PacBio CCS/HiFi sequences both strands of a DNA molecule".    I don't know the chemistry or really anything about 6mA modifications,  if its impossible no problem then.

jrobinso:
It always takes me some days to get back into this, I think I'm good now.

What's messing up the coverage track is a built in assumption that only ~1/2 the reads are informative for base calling.  The 5mC data I've seen certainly looks like that, but I learn from @ctsa this is a convention not reality.     So the other 1/2 of the reads (those on the "wrong" strand) are not informative and should not be considered when estimating the % of reads modified for the coverage track.     I now need to modify that, its incorrect here.   I think it should be safe to assume that if only the "+"  form of a modification is included as a tag (e.g. C+m) that the single-strand assumption should apply,  if its complement is also present then all the reads are potentially informative.      But this is a separate more general issue than just "6mA".

jrobinso:
@ctsa I'm working now on generalizing the 6ma option so we don't have to do this again for every modification.   Its straightforward.  However I think this first test (for the base) should not be neccessary, the "a" modification (6mA) only applies to base "A",  so we shouldn't need to test the canonical base, there are no other values possible (other than "A" or "T").   No response necessary, just letting you know I'm dropping it.

```
 if (bmSet.getCanonicalBase() != 'A' && bmSet.getCanonicalBase() != 'T') continue;
                            if (! bmSet.getModification().equals("a")) continue;
```

jrobinso:
@ctsa @mrvollger Forgive me for extending this conversation, but I am a bit confused about the difference in treatment between the 5mC and 6mA calls in this dataset.   @ctsa alluded to this above but I don't get it.   I'm trying to do the right thing for the coverage track, or at least a consistent thing between the 2 calls.

In this dataset there is no G-m calls corresponding to the C+m calls.   So reads on the bottom strand have no "m" calls (see screenshot).   What happened to those calls (G-), presumably they were made as the T- calls were.   More importantly what should the coverage reflect.   If at a "C" site  100% of the top strand alignments were called 5mC with 100% confidence I think one would expect a solid red bar.   The bottom strand is not informative.   This is not true of the "A" sites,  as both strands are informative for the modfication.   Do you see the difficulty?

Screen Shot 2023-06-19 at 1 37 10 PM

mrvollger:
Hi @jrobinso,

I have also been confused about how CpG is encoded in the past and @ctsa should confirm all of the following... but my understanding is that the 5mC encoding is not done totally correctly, or has been abbreviated to omit the minus strand. I think the reason for this is that if you have a 5mC on the + strand it is almost always also present on the - strand as well so encoding both ends up being largely redundant, and the other reason I think, is that PacBio uses the joint information across both the + and - strand to calculate that 5mC confidence. 

However, as I said @ctsa should confirm all this since I am also slightly confused about the encoding at times. 

Looking at your figure I think you have the "correct" coverage for 6mA but 5mC might need special treatment in order to sync with previous conventions.

jrobinso:
Thanks @mrvollger but I don't follow the explanation.    There's too many "strands" involved.    Every molecule of double-stranded DNA will of course have complementary C/T,  but the modification can only be on the C.   There's no difference really in 5mC and 6mA modifications in this regard.   The CpG consideration is really not relevant that I can see, it is with bisulfite sequencing but base modification detection is one base at a time, in my understanding, it doesn't matter what is happening on the neighboring base.

I am probably not communicating clearly, but I think we can agree some special treatment is needed for the 5mC encoded data.   I am working on that, but in a general way.   Any dataset that lacks complementary calls  will receive the same treatment,  e.g if we only had A+a and no corresponding T-a.   My idea is to remove counts of the complementary base as non-informative,  with no impact on colored bar height.   5mC mode already does this,  but I don't like the implementation.

ctsa:
The community has settled on a 5mC representation that's out of spec, some of the reasons:
1. No good way to represent the joint C+/G- likelihood if following the spec representation
2. Simpler representation in the MM/ML tags...
3. Inertia. There's not much need to distinguish single-strand/hemimethylated sites in human to begin with, so many tools write just the C+ side and likewise many parsers only read this side as well.

Regarding how to handle in IGV, I would start by recognizing that it is an out-of-spec convention. Maybe there could be a checkbox somewhere in the config to say the equivalent of "'C+' interpreted as fwd(5mCpG)/rev(5mCpG)"? In this mode, the coverage of the C should sum over the C+ and G- at the following position. (probably in this mode the coverage for the G at the following position should do likewise).

Perhaps this creates a giant mess inside IGV though -- if the coverage column calculation now needs to work with adjacent columns. It might be okay to leave as-is and document the situation.

jrobinso:
@ctsa. Is a 5mC modification exclusive to CpG sites?  Actually I don't think it is,  in human maybe but in other organisms there are other contexts (e.g. CHH, CHG, etc).  Perhaps experiments for those organisms and contexts don't use this convention and the defaults would work.  Do you know?

I don't understand point 1.,  do you mean joint likelihood of a C+ and its adjacent (wrt genomic position) G-.   Are these calls made jointly?  So its really a joint dinculeotide call?   So these calls are always in pairs, I will never see a 5mC call at some "C" site without a "G" following?

Treating the C and G of a CpG site as a unit would certainly work, this is what the bisulfite mode does it you select CG as the context.   So that could be an option for treating 5mC calls, certainly most people who care are going to be working in human.   However this came up early on,  and I thought we decided we don't want to treat the CpG site as a unit.

jrobinso:
For modifications other than 5mC will we always expect complementary tags (e.g. A+a/T-a)?  

I'm wondering if a lone C+m with no G-m is a strong indicator of this convention for CpG.    That might be useful to know.   Do you know if ONT experiments use this same convention?

jrobinso:
Treating adjacent C+ calls at CpG sites as a unit is going to be a mess, and implicitly assumes an out-of-spec convention that might not be followed everywhere and at all times.    I'm pursing a different approach that should mostly work visually at least.    In the case of a "non-matched" mod key, e.g. C+m but no T-m,  only the informative alignments are going to impact the height of the coverage bar.   So at a "C" reference position the height will represent the likelihood of a modification reflected in the + strand alignments.   At a "G" reference the likelihood reflected in the - strand alignments.   The other strand is not informative so it will just be ignored.    The only complication is accounting for snps, but its just bookkeeping.

The actual height in all cases, BTW,  reflects the sum of likelihoods,  not just counts,  so a pileup of low likelihood calls is relected in a shorter colored bar height than a pileup of high likelihood calls.     This is how it's coded now, an alternative would be to let the height reflect the number of calls and alpha the average likelihood.   There's more information but is messy to look at.   Using the likelihood sums is directly analogous to snp treatment, where the height is proportional to the sum of base qualities.

ctsa:
> @ctsa. Is a 5mC modification exclusive to CpG sites? Actually I don't think it is, in human maybe but in other organisms there are other contexts (e.g. CHH, CHG, etc). Perhaps experiments for those organisms and contexts don't use this convention and the defaults would work. Do you know?

Correct that it is not exclusive to CpG. I believe endogenous 5mC only occurs at CpG contexts in humans. Any 5mC calls in more general contexts like CHH and CHG will probably be represented in the MM tags according to spec, and you would find these calls more like the 6mA calls. The 5mCpG convention is the only community exception I know about.

> I don't understand point 1., do you mean joint likelihood of a C+ and its adjacent (wrt genomic position) G-. Are these calls made jointly? So its really a joint dinculeotide call? So these calls are always in pairs, I will never see a 5mC call at some "C" site without a "G" following?

These particular calls that we're referring to in the example bam from Mitchell are called jointly over the CpG dinucleotide, under the assumption that the hemi-methylated state is highly improbable. I can't speak for the exact representation of other callers.

ctsa:
> For modifications other than 5mC will we always expect complementary tags (e.g. A+a/T-a)?

Even for 5mC calls we expect the complementary tags (C+m/G-m) if we were more closely matching the spec.

ctsa:
> Treating adjacent C+ calls at CpG sites as a unit is going to be a mess, and implicitly assumes an out-of-spec convention that might not be followed everywhere and at all times. I'm pursing a different approach that should mostly work visually at least. In the case of a "non-matched" mod key, e.g. C+m but no T-m, only the informative alignments are going to impact the height of the coverage bar. So at a "C" reference position the height will represent the likelihood of a modification reflected in the + strand alignments. At a "G" reference the likelihood reflected in the - strand alignments. The other strand is not informative so it will just be ignored. The only complication is accounting for snps, but its just bookkeeping.
> 
> The actual height in all cases, BTW, reflects the sum of likelihoods, not just counts, so a pileup of low likelihood calls is relected in a shorter colored bar height than a pileup of high likelihood calls. This is how it's coded now, an alternative would be to let the height reflect the number of calls and alpha the average likelihood. There's more information but is messy to look at. Using the likelihood sums is directly analogous to snp treatment, where the height is proportional to the sum of base qualities.

I like the idea that, so long as the representation and community conventions are messy, you would not bend over backwards to try to parse out off-spec conventions, so this sounds sensible.

When there's another opportunity I'll try to join into the bam spec discussion to see what's possible for dinuc calls.

jrobinso:
OK,  thanks for the clarification its clear now.  I think you nailed it above,  the spec does not accomodate dinuc calls, something like. 'CG+m' and 'CG-m' is needed.

mrvollger:
So I think this is due to PacBio ccs sequencing. Both strands are sequenced many times in a rolling circle for as long as the polymerase can. But this means that often one strand has 1 more pass than the other strand because of when the polymerase falls off. This gives one strand more accuracy relative to the other. And I think that PacBio convention is to make the plus strand the one with more passes but @ctsa would have to confirm all this.

jrobinso:
Ok, just noting it, in this small sample there is a definite bias towards detecting modifications in the plus strand, beyond any accuracy considerations.   This is an advantage I guess in visualizing each nucleotide as this bias is obvious, whereas treating the CG as a unit would not show it.   But then again this is just a small sample dataset, just noting it,  at first I thought it was an IGV bug.

ctsa:
> @ctsa @mrvollger An observation on this test dataset -- it appears that the average likelihood of an "m" modification on the negative strand alignments ("G" sites) is consistently lower than the corresponding positive strand alignments ("C" sites). I wouldn't expect this systematic difference.

Hi Jim, It sounds like you're talking about reads aligned to the fwd or reverse strand with respect to a reference sequence here, in which case I wouldn't expect any bias. All of the base modifications are applied prior to alignment. For the case of 5mC, as we've discussed, all of the base mod annotations are registered to the fwd-strand of the CCS template, and CCS template rev strand base mods are implicit.

ctsa:
@jrobinso - Since there's a lot of detail to followup on I'd be happy to discuss it in a call if you'd like.

mrvollger:
I'd be interested in joining this convo if it happens to make sure I am doing things right with my tools as well.

jrobinso:
I think IGV is all sorted but it would be useful to just talk through what's being done as a review and to be sure we're all on the same page.   I'm available this afternoon after 2 and late afternoon (> 3) tomorrow.

ctsa:
Sounds like there's interest so let's see if we can coordinate something, perhaps tomorrow, I'll pick this up on email.

PR #1354, opened 2023-07-05
Added nucleotide color options to the preferences menu

KevinCLydon:
Added the ability to set nucleotide color options to the preferences menu, under General > Nucleotide Colors.

Also modified SequenceRenderer to update nucleotide colors from preferences when a new SequenceRenderer is instantiated.  This way, the colors will be updated when reloading the session or loading a new session.  I'm not sure if this is the ideal solution to the problem of updating the colors without having to restart, so I'd be very happy to change it if there is a better way.


jrobinso:
Thanks @KevinCLydon .   Re the colors,  I think the simplest thing would be to just fetch the colors on each draw, I can't imagine we save much by caching these colors.  I can make this change.

Issue #1271, opened 2022-12-19
Genome switching via port commands does not always work correctly


ubuntolog:
Hello Jim,

I have encountered a strange problem while switching between genomes by means of port commands.

We have two instances of IGV running at the same time on different ports. Each of them is initialized and started via port commands. The instances use different genomes. Very often the second instance tries to load a genome and fails, after that it switches to the genome used by the first instance. Sometimes when the second instance is loaded with the correct genome IGV duplicates the Sequence track (see the screenshot):

![image](https://user-images.githubusercontent.com/5293569/208402860-b3078856-130c-40d2-9b51-891a36d6ad16.png)

I suspect that it may have something to do with the "DEFAULT_GENOME_KEY" from the settings. IGV overwrites the settings file each time we load a genome via port commands and switch genomes (also via port commands).

In the log file I often see a severe error message for the second genome saying: "[ReferenceFrame] Null chromosome:"

As far as I understood from the source code, each port command is running in a separate thread (but I did not dig too deep). Is it possible that the genome thread is too slow, but IGV does not care about it and loads all the rest tracks? I also tried to use "setSleepInterval" command to introduce an artificial delay before loading a genome and it improved the situation. However, IGV still does not work reliably for our scenario with two instances.

The problem is very difficult reproduce and I could not come up with a better description. We also believe that it may be related to the bug we reported earlier: https://github.com/igvteam/igv/issues/1094

I will try to debug it and will let you know, if I find out more.

jrobinso:
You say "very often the second instance tries to load a genome and
fails...".      Any idea why it fails?    If it fails DEFAULT_GENOME_KEY
might come into play, but not before.

ubuntolog:
Thank you for replying so quickly. It is hard to be very specific, since the problem does not occur consistently (or I have not found the pattern yet). By saying "very often" I mean in 80-90% of cases. I assume that simultaneous writing/reading to/from the same config file (by different IGV instances) maybe causing that. It seems to me that modifying the config after calling the "genome" port command is not a good idea. Maybe the modification of the settings should be made only from the GUI.

jrobinso:
I'm not sure why this is still open, commit a41b0296 should fix it.   If the problem recurs reopen with specifics.

Issue #1094, opened 2022-01-27
Command sent via port is sometimes ignored


marc-sturm:
Hi,

I noticed a bug in IGV 2.11.9 under Windows10 that is related to commands via the port.
Our application starts IGV using a custom port via the `--port` argument.

After IGV is started, we wait until a connection to the port can be established.
Then we send the `new` command.

However this command is sometimes ignored and no reponse is sent.
I think this happens when the command is send while the "Loading genome..." message is still shown.
It seems that the port only works after the genome has been loaded and should not be enabled before that time.

Best,
  Marc

jrobinso:
OK I'll look at it.  BTW the "new" command will not do anything after a fresh load in any event, what is it you expect it to do?  Are you effectively using it as a "ping"?  I could add a do-nothing ping type command for this purpose.

jrobinso:
The command listener has to be started before the initial genome load, this was to allow a specific case of setting credentials through the port for a privately hosted genome.   Its true that some commands, such as "new", will fail if there is no genome loaded.    I can look into why those commands are not returning an error message, but will that address your problem? Perhaps we need a command that returns OK after the genome is loaded.

jrobinso:
Another possibility is to just hold commands in a queue until the initial genome is loaded, then execute them in the sequence received.   I will look into that option.

jrobinso:
@marc-sturm Try specifying your genome as the first command, instead of new.  If "genome" is the first command it will block until its loaded.

PR #1360, opened 2023-07-18
Adds a port command for getting information about the currently selected genome

ubuntolog:
The new "currentGenome" port command will return the path of your currently selected and loaded genome. This port command can be used as a fix (or a good workaround) for the issues we have previously created:

https://github.com/igvteam/igv/issues/1271
https://github.com/igvteam/igv/issues/1094

Other users may also find it helpful while using port commands from their applications. When switching between multiple genomes, one may need to know which genome is currently selected.


jrobinso:
Looks good, but I suggest the name be "currentGenomePath",  to be clear its not returning the ID.

ubuntolog:
> Looks good, but I suggest the name be "currentGenomePath", to be clear its not returning the ID.

yes, it sound better, I will adjust my code

ubuntolog:
I have adjusted the naming

jrobinso:
Ok, merged.

ubuntolog:
Thank you!

PR #1361, opened 2023-07-18
Added a check for identical tracks

ubuntolog:
In the issue https://github.com/igvteam/igv/issues/1271 I have mentioned the situation when IGV creates two identical tracks. I have added a simple check that should eliminate the problem.

### DIFF
--- a/src/main/java/org/broad/igv/track/TrackGroup.java
+++ b/src/main/java/org/broad/igv/track/TrackGroup.java
@@ -115,7 +115,15 @@ public class TrackGroup {
         if (track == null) {
             log.warn("Attempt to add null track");
         } else {
-            tracks.add(track);
+            boolean alreadyExists = false;
+            for(Track t : tracks) {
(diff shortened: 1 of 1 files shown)


jrobinso:
Its perfectly legal to have multiple tracks with the same name,  #1271 is specifically focused on the sequence track which is a special case.   I thought that had been fixed,  but I might be thinking of igv.js.   Is the duplicate sequence track still an issue with the latest release?

ubuntolog:
> Its perfectly legal to have multiple tracks with the same name, #1271 is specifically focused on the sequence track which is a special case. I thought that had been fixed, but I might be thinking of igv.js. Is the duplicate sequence track still an issue with the latest release?

I will check which version we have now "in production" and get back to you

ubuntolog:
The bug still persists in the latest release version
image

jrobinso:
OK I will fix it for the next release (2.17.0).

ubuntolog:
Thank you.

PR #1362, opened 2023-07-18
Set focus on OK button in LoginDialog

ubuntolog:
A small change that sets a focus on OK button for the login dialog (e.g. if a proxy server is used). Currently if a user types in credentials and hits Enter, nothing happens. After incorporating this tiny fix, OK button will be pressed on hitting Enter.


jrobinso:
Thanks

PR #1368, opened 2023-08-07
Added empty line to 204 http response

BOBAHDEP:
For case when 'result' is null and IGV sends HTTP_NO_RESPONSE http headers are not followed by empty line (but it should be so https://httpcodes.fyi/204/). It results in missing Access-Control-Allow-Origin header in response for 204 code.

Added empty line to http response in case result is null.


jrobinso:
Good catch, thanks!

PR #1413, opened 2023-10-19
Upate to htsjdk 4.0.2

lbergelson:
Htsjdk added a new NamedFeature interface that is clashing with IGV's slightly different NamedFeature. Renamed NamedFeature -> IGVNamedFeature
Slight change to IGVFeatureReader interface to make it return a CloseableIterator so that it can be handled with try-with-resources.

It might be worth updating the new Htsjdk interface to include the displayName() method as well so that can just be used directly?

### DIFF
--- a/src/main/java/org/broad/igv/feature/tribble/TribbleReaderWrapper.java
+++ b/src/main/java/org/broad/igv/feature/tribble/TribbleReaderWrapper.java
@@ -25,6 +25,7 @@
 
 package org.broad.igv.feature.tribble;
 
+import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.tribble.*;
 import htsjdk.tribble.index.Index;
 import htsjdk.tribble.index.IndexFactory;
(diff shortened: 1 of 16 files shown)


lbergelson:
This doesn't set the new environment variable yet.

jrobinso:
The entire "Feature" class/interface structure could use a look, with a view to remove duplicate interfaces wrt the htsjdk.   Not for this PR.

lbergelson:
Rebased this on master, hopefully the test pass now, it looks like it was some sort of transient issue contacting blast.

src/main/java/org/broad/igv/track/TribbleFeatureSource.java
```diff
@@ -332,45 +333,30 @@ private NonIndexedFeatureSource(AbstractFeatureReader basicReader, FeatureCodec
...
 
                     String seqName = f.getChr();
                     String igvChr = genome == null ? seqName : genome.getCanonicalChrName(seqName);
 
-                    List<Feature> featureList = featureMap.get(igvChr);
-                    if (featureList == null) {
-                        featureList = new ArrayList();
-                        featureMap.put(igvChr, featureList);
-                    }
+                    List<Feature> featureList = featureMap.computeIfAbsent(igvChr, k -> new ArrayList<>());
```
lbergelson, line 345: I did a bit of refactoring in this method, every where else is basically just renames.

src/main/java/org/broad/igv/feature/tribble/TribbleReaderWrapper.java
```diff
@@ -69,13 +68,11 @@ public synchronized Iterator<Feature> query(String chr, int start, int end) thro
...
             }
             return featureList.iterator();
-        } finally {
-            if(iter != null) iter.close();
         }
     }
 
     @Override
-    public Iterator<Feature> iterator() throws IOException {
+    public CloseableIterator<Feature> iterator() throws IOException {
```
lbergelson, line 75: This is autocloseable now so it can be handle automatically.

PR #1427, opened 2023-10-31
Event code cleanup

lbergelson:
I started to take a look at the event code a few weeks ago.  I did a few things, none of them are huge changes but I think it makes them easier to use/discover and removes some boilerplate.   It doesn't solve the problem of not knowing what they are meant to mean but it makes it a bit easier to discover the possible events and removes a bunch of them that were never used.  Documenting what events should mean / is probably something worth looking at in the future.

    * Cleanup some of the event code by taking advantage of newer java 17 features
      - make many event types records and delete some unused events / fields
      - fix some intellij complaints in event usage / IGVEventBus
    * cleanup ViewChange constructors
    * make all events be part of a sealed interface.  This doesn't have a real functional effect right now but it makes it easier to list / discover all the existing events that are used so there's less chance of duplicating functionality.  
    
Also as part of looking at this I found and fixed a bug where we were repainting AlignmentTrack twice in a row because of a missed `break` statement.


src/main/java/org/broad/igv/sam/AlignmentTrack.java
```diff
@@ -367,30 +368,22 @@ public void init() {
...
+            switch (e.type()) {
+                case ALLELE_THRESHOLD -> dataManager.alleleThresholdChanged();
+                case RELOAD -> {
                     clearCaches();
                     repaint();
-                case REFRESH:
-                    repaint();
-                    break;
+                }
+                case REFRESH -> repaint();
```
lbergelson, line 383: There was a silly bug here where on a RELOAD we would repaint twice since it would fall through without a `break`

PR #1431, opened 2023-11-14
Set develop flag to true for develop builds

lbergelson:
It seems like this should be set to true in the default properties.  It appears that it's set to true in the released bundled builds also ( which is weird). 

It might also make sense to just completely remove the isDevelop() flag from the code since it's rarely used and is very likely to cause confusion (as it did in #1430 ).


jrobinso:
@lbergelson I think we should remove this flag.

jrobinso:
I've removed the property

PR #1437, opened 2023-11-16
Fixes the problem of multiple tokens for identical URL patterns

ubuntolog:
I have recently come across a very strange problem while using SetAccessToken port command. From my application I was updating an access token, however, IGV kept using the old token. The problem could not be reproduced consistently. Fortunately, I was able to track it down to accessTokens HashMap. The variable uses a Pattern object as a key, which allows to have an arbitrary number of identical URL patterns for different tokens (still valid and already expired ones). Since HashMap is not sorted, sometimes it may return correct tokens, but very often it will give you expired ones. Expired tokens are stored indefinitely and are not overwritten by the new ones.
This pull request offers a solution to the described problem by introducing another HashMap and by storing the Pattern object as a value.

### DIFF
--- a/src/main/java/org/broad/igv/util/HttpUtils.java
+++ b/src/main/java/org/broad/igv/util/HttpUtils.java
@@ -96,8 +96,11 @@ public class HttpUtils {
     private final int DEFAULT_REDIRECT_EXPIRATION_MIN = 15;
     private Map<URL, CachedRedirect> redirectCache = new HashMap<URL, CachedRedirect>();
 
-    // oauth tokens set from command line script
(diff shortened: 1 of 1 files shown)


ubuntolog:
Your solution seems to be a more elegant alternative. According to the documentation (I never used Deque), it should do the job. A queue is a better data structure for this task, code is more readable, and we do not need an extra variable. I will test it locally and let you know, if I find problems. Thank you

ubuntolog:
Changed from https://github.com/igvteam/igv/pull/1453 seem to work as expected.

Issue #1468, opened 2024-01-13
v2.17 in windows can't get read information by left-click


sqwwww:
v2.17 in windows can't get read information by left-click,  maybe there was a bug in it? please let me know if there is any information.

jrobinso:
I can't reproduce your problem, you will have to give us more details.

huangnengCSU:
Hi.
Same problem of v2.17, I can not get the alignment information when left-click one nucleotide of the read. I have tested on MacOS and window11.

To reproduce the problem. First, you show all base of the read. In the previous version, when you  click one nucleotide then you can get the alignment information of the read including Read name, Read length, Mapping quality and base quality of the nucleotide as well. In the latest version, when you click the nucleotide, you will get nothing. Even if you get alignment information by chance, the base quality is not in it anymore.

jrobinso:
Hi,  sorry but I still can't reproduce your problem (see screenshot).   This might be specific to your dataset.    Could you zip and attach your igv.log file?  This might give a clue to the problem.   You can find it in an "igv" folder in your home directory, the filename is usually igv0.log.
<img width="547" alt="Screen Shot 2024-01-21 at 12 46 36 PM" src="https://github.com/igvteam/igv/assets/933148/ac194587-c896-402b-9cd3-fd7a5e13d3f0">

PR #1471, opened 2024-01-23
isFlow robustness, common code into (new) FlowUtil

dror27:


src/main/java/org/broad/igv/ultima/FlowUtil.java
```diff
@@ -0,0 +1,69 @@
...
+
+/*
+ * misc utilities to handle flow base (Ultima) data
+ */
+public class FlowUtil {
+
+    public static final String TAG_T0 = "t0";
+    public static final double MIN_PROB_DEFAULT = 0.01;
+    public static String ATTR_TP = "tp";
+    public static String ATTR_TI = "ti";
```
ilyasoifer: @dror27 - can we get rid of `ATTR_TI` please? Only very old data used that tag

ilyasoifer (commented):
@dror27 thanks! Just a minor request, will let @jrobinso make a more thorough review

ilyasoifer:
Fixes #1468

PR #1474, opened 2024-01-25
Catch errors parsing supplementary alignments

jrobinso:
Catch and log errors parsing SA tag to annotate clipped ends.  This allows alignment rendering to continue in case of malformed SA tags.


lbergelson:
@jrobinso  This looks good to me.

lbergelson (approved):

Issue #1394, opened 2023-09-20
(For long read RNAseq) Is there any way to force IGV to display 1 read per line?


cbohrson:
I'm interested in doing this because in some cases this might improve visualization of complex transcript isoforms.

I searched the web and was unable to find a solution.  Also tried messing with IGV settings; only thing I could manage was assigning a new tag to each read and then grouping by tag (using IGV 2.16.2).

This didn't really work as with the reads grouped many other visualization options became unavailable.  Also the dashed lines dividing the groups interfered with the view.

Is there any other solution?

Thanks

jrobinso:
Hi, I'm not really sure what is being requested here.  To be precise IGV displays alignments not reads.    Are your reads represented by multiple alignments (i.e. so-called chimeric, supplementary, or split alignments)?

f we literally displayed 1 alignment per line we would need millions of lines in some cases.   I can't quite get my head around this, an annotated screenshot might help.

maximilianh:
I think UCSC does this and calls it "full" mode, one single alignment per
line on the screen. It's funny that we just discussed this view mode and
thought that probably no one uses or wants it...  :-)

jrobinso:
@cbohrson @maximilianh The thing that's never been clear to me is what should happen when you pan left or right in full mode.

jrobinso:
This might work, alignments in view are stacked based on start position.  As new alignments come into view the stack is reshuffled as needed.  This doesn't happen while actively panning (dragging), rather when the mouse is released.  This will be unsuable for short reads,  where there might be many 10s of thousands of alignments in view,  but might be o.k. for long ones, although really zoomed out views could still be a problem.   Of course you can switch to another display mode then.

dpaudel-tb:
@jrobinso I was wondering if there were any updates on this? Having "one single alignment per line on the screen" would be very helpful specially with long reads. One use case is to visualize full length reads vs partial reads. This enhancement would be very elegant. For full lengths, current implementation works but for partials I do not want multiple reads visualized in the "same line".

jrobinso:
This is difficult to implement in IGV.   You will have 1 line per alignment loaded, which can be 100s of thousands if not more.    There might be a solution but it will involve some redesign, it is not a quick fix.   As suggested by @maximilianh above you can use the UCSC browser for this.     Leave the issue open for now but I can't give an ETA.

jrobinso:
[...]

PR #1531, opened 2024-07-02
Support full mode for alignment tracks

jrobinso:


jrobinso:
Support "full" mode (1 alignment per row) for alignment tracks.  Fixes #1394

PR #1542, opened 2024-08-14
Adding additional options for duplicate reads.

lbergelson:
This is in response to a feature request from @rickymagner.  It doesn't seem like there was any way to see which reads are duplicates.  Added some new options to help.

* new Group By Duplicates option
* new Duplicates submenu which allows rapid toggling of duplicates between three states
  * Filtered: not visible
  * Visible: treated like normal reads
  * Textures: rendered with a visible slash texture

Showing the new grouping and cross hatching
Screenshot 2024-08-14 at 3 17 20 PM

The new menu option
Screenshot 2024-08-14 at 3 17 28 PM


lbergelson:
huh, I seem to have an test failure... I'll look into that...

lbergelson:
Fixed. I made the faulty assumption that renderOptions wasn't allowed to be null there.

jrobinso:
I love the feature.   I wonder if the options should be verbs, i.e "filter", "show", and "texture".   I guess texture isn't a verb, can I texture something?   Or space isn't a problem, we could be explict,  "filter duplicates", "show duplicates", and "texture duplicates".   Again I'm not sure texture is really a verb,  but its close enough in context.

rickymagner:
Apparently it can be used as a [verb](https://www.dictionary.com/browse/texture) as well. I like making them verbs since it feels more like an action you're performing in the app.

src/main/java/org/broad/igv/sam/AlignmentPacker.java
```diff
@@ -438,69 +438,64 @@ private Object getGroupValue(Alignment al, AlignmentTrack.RenderOptions renderOp
...
-            case LIBRARY:
-                return al.getLibrary();
-            case READ_GROUP:
-                return al.getReadGroup();
-            case LINKED:
-                return (al instanceof LinkedAlignment) ? "Linked" : "";
-            case PHASE:
-                return al.getAttribute("HP");
-            case TAG:
+        return switch (groupBy) {
```
jrobinso, line 441: This is a big improvement.

jrobinso (approved):

PR #1543, opened 2024-08-14
Move the amazon credentials check at startup into an async thread

lbergelson:
I was checking to make sure I hadn't slowed down rendering by adding more stuff to it, I noticed that loading the amazon credentials was adding about half a second to the startup time.   This should make the check asynchronous.  Could you make sure it seems sane to you?  I don't have amazon credentials configured so it's hard for me to check if I did something weird to them.

### DIFF
--- a/src/main/java/org/broad/igv/ui/IGVMenuBar.java
+++ b/src/main/java/org/broad/igv/ui/IGVMenuBar.java
@@ -211,7 +211,7 @@ public class IGVMenuBar extends JMenuBar implements IGVEventObserver {
     }
 
     public void updateAWSMenu() {
-        AWSMenu.setVisible(AmazonUtils.isAwsProviderPresent());
+        SwingUtilities.invokeLater(() -> AWSMenu.setVisible(AmazonUtils.isAwsProviderPresent()));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jrobinso:
As you aren't even loading credentials I'm not sure why it should take any time.    I'll investigate.

jrobinso:
Oh wow, it takes a long time if you don't have AWS credentials setup, as you discovered.  2 seconds for me.  51 ms with my aws credentials.   This is a long delay, good catch.

jrobinso:
Looks good, but setVisible should be called from the AWT event thread.   As its now going to be called from the runnable thread we need to use SwingUtilities.invokeLater().   I will make this change.

jrobinso:
The costly part of this, BTW,  is something in the AWS SDK that walks a provider chain looking for credentials.   If you have them setup it finds them right away.

lbergelson:
Thanks for the fix.  I wasn't sure if you had to call that from the event thread or if it was ok to do from anywhere.

jrobinso:
Its not always clear which methods have to be called from the event thread,  I think there is a master list somewhere.  In practice there's almost never a problem either way.

PR #1568, opened 2024-09-12
Update to Java 21 and gradle 8.10.1

lbergelson:
* Move from Java 17 -> Java 21
* Update from gradle 8.0.2 -> 8.10.1
  * Replace some deprecated gradle code with newer styles
  * Something changed in the module resolution with this update which caused us to not be able to resolve automatic modules.
  * Introduced a new gradle plugin which fixes the problem and removed the manual module path manipulation we were doing.


lbergelson:
I forgot to update the github actions, but now it's working.

PR #1571, opened 2024-09-16
Remove snappy

lbergelson:
This should remove snappy which will hopefully fix our notarization issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,6 +104,8 @@ configurations {
         exclude group: 'xalan', module: 'serializer'
         exclude group: 'xalan', module: 'xalan'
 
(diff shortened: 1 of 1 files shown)


lbergelson:
A made a change in htsjdk but I need to do a release before we can fix this branch. https://github.com/samtools/htsjdk/pull/1715

build.gradle
```diff
@@ -359,7 +361,8 @@ tasks.register('createMacAppWithJavaDistZip', Zip) {
     }
 }
 
-task createWinDist(type: Copy, dependsOn: createDist) {
+tasks.register('createWinDist', Copy) {
```
lbergelson, line 364: I missed these in the 21 update branch.

PR #1580, opened 2024-09-24
Correct rendered range of symbolic VCF alleles

ctsa:
This patch adjusts the start position of symbolic VCF alleles to better match the VCF spec definition:

https://github.com/samtools/hts-specs/blob/4cde0e235b4e4cfbd67a0c3a38abea611d56d256/VCFv4.1.tex#L165

Motivated by discussion on https://github.com/samtools/hts-specs/issues/792  -- cc: @jrobinso / @d-cameron

### DIFF
--- a/src/main/java/org/broad/igv/variant/vcf/VCFVariant.java
+++ b/src/main/java/org/broad/igv/variant/vcf/VCFVariant.java
@@ -385,6 +385,17 @@ public class VCFVariant implements Variant {
         if (variantContext.getType() == VariantContext.Type.INDEL || variantContext.getType() == VariantContext.Type.MIXED) {
             prefixLength = findCommonPrefixLength();
         }
+
+        /**
+         * Since v4.1, the VCF spec has defined POS as the base preceding the polymorphism for any symbolic allele.
+         *
+         * https://github.com/samtools/hts-specs/blob/4cde0e235b4e4cfbd67a0c3a38abea611d56d256/VCFv4.1.tex#L165
+         *
+         */
+        if (variantContext.getType() == VariantContext.Type.SYMBOLIC) {
+            prefixLength = 1;
+        }
+
         this.start = (variantContext.getStart() - 1) + prefixLength;
     }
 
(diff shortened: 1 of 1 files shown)


jrobinso:
I'm sure this is correct but I'll ask @lbergelson to look, if for no other reason so he is aware.

ctsa:
Thanks @jrobinso, the more eyes on the better.

src/main/java/org/broad/igv/variant/vcf/VCFVariant.java
```diff
@@ -385,6 +385,17 @@ private void calcStart() {
...
             prefixLength = findCommonPrefixLength();
         }
+
+        /**
+         * Since v4.1, the VCF spec has defined POS as the base preceding the polymorphism for any symbolic allele.
+         *
+         * https://github.com/samtools/hts-specs/blob/4cde0e235b4e4cfbd67a0c3a38abea611d56d256/VCFv4.1.tex#L165
+         *
+         */
+        if (variantContext.getType() == VariantContext.Type.SYMBOLIC) {
```
lbergelson: I think this makes sense in general.   I think we want to be calling getType with ignoreRef == true.  That accounts for the case where you're loading a GVCF with `<NON_REF>  or <*>` alleles.    There is a special exception for those even though they're considered symbolic.  

>  The exception to this is the <*> symbolic allele for which the reference call interval includes the POS base.

https://github.com/samtools/hts-specs/blob/4cde0e235b4e4cfbd67a0c3a38abea611d56d256/VCFv4.4.tex#L357

I suspect not all tools correctly handle this for all cases also so it may cause some confusion.  I'm thinking in particular of anyone who might be using things like symbolic IUPAC bases (`<R>` etc) which are almost certainly treated as though they were a single SNP base without any padding.   I haven't actually run into those in the wild though, so maybe they're only a theoretical issue.

```suggestion
        if (variantContext.getType(true) == VariantContext.Type.SYMBOLIC) {
```

lbergelson: There's probably also something to be done at the genotype level to strip padding from mixed SYMBOLIC /  not symbolic allele mixes.

ctsa: Thanks for pointing this out @lbergelson, I missed that there's a non-ref exception in the spec. I also didn't want to dig into the MIXED symbolic allele cases because these could get more complicated and I'm mainly motivated to correct this for the most common single-allele SV cases.

lbergelson (approved):
Looks good to me.  Thank you @ctsa

PR #1590, opened 2024-10-01
Fix issue1512

kunmonster:
This commit is for fixing the issue  #1512 with the solution from [jrobinso in #1588](https://github.com/igvteam/igv/pull/1588#issuecomment-2380423738)  , the changes are as follows.

1. Get the configurations of all screens.
2. Check whether user preference is null.
2. Check whether (x, y)  in user preference is within a certain screen.

fix
- the case #1512 in windows.

other behaviors
- when the window of igv was  crossing two screens in windows, the previous code will open it in the  same  position as last closing, this code will change it to the upper left corner of the screen containing the (x,y) of IGV
- work normal in windows 11 with jdk21 or jdk17.

Please review the code and test it in the macos.


jrobinso:
Nice.  Sorry for the delay.

kunmonster:
> Nice.  Sorry for the delay.

Not matter.Thank you.

Issue #1598, opened 2024-10-08
Problem using AWS Cognito to authenticate users for S3 access


mblanche:
Hi there, 

I tried following the steps described on the [UMCCR Genomics Platform Group](https://umccr.org/blog/igv-amazon-backend-setup/) to use AWS Cognito user pools and identity pools to authenticate access to S3 resources containing files to be used in IGV. The steps describe could benefit from a refresh as the AWS console for Cognito is dramatically different from when the tutorial was writen. Nonetheless, I was able to create both a user and identity pool as well as assigning a role with S3 access but somehow, the callback to IGV after the login goes to a web page that says, "This page is not working" and IGV does not seem to be downloading any content from S3.

Just to recap the step I did:
1. I created a user pool with the following attributes:
a. I have create an app client with a client id and client secret
b. The app client is set to auth flows ALLOW_REFRESH_TOKEN_AUTH and ALLOW_USER_SRP_AUTH
c. I have a hosted UI with a callback set to http://localhost:60151/oauthCallback
d. Hosted UI identity provider set to Cognito user pool
e. Hosted UI OAuth 2.0 set to Authorization code grant
f. Hosted UI OIDC scopes set to aws.cognito.signin.user.admin, Email, Ohone, Profile and OpenID
2. Then created an Identity pool selecting "Authenticated access" and "Amazon Cognito user pool"
3. Creating a new IAM role with name igv-cognito-access-role
4. Selecting the user pool id and App Client Id created in 1. as well as selecting "Use default authenticated role" and "Claim mapping inactive", then naming the user pool and accepting the settings
5. I then migrated to the User pool IAM role and added the policies, as in-line polices, described on [UMCCR web page](https://umccr.org/blog/igv-amazon-backend-setup/)
6. Created an S3 webserver to host the oauth-config.json (I'll put the content later)
7. Updated the IGV advance pref to point to the url of the hosted config file

When I open IGV, I get the Amazon menu, selecting the login trigger the Cognito hosted UI where I can manage my credentials, etc... Then when I login I get redirected to an error page saying "This page is not working" and nothing gets loaded from S3 into IGV.

Not sure what I am missing? Let me paste the different configs and log, no worries about the different secrets, everything got wiped out cleaned...

Let me know if there's more info required, I really need this to work for the project I'm working on right now.

Thanks

Marco

## Config file hosted on static S3 web server

**oauth-config.json**
```
{
  "apiKey": "",
  "project_id": "igv",
[...]

PR #1599, opened 2024-10-11
Fix for Amazon failing.

lbergelson:
* Fix the amazon failures due to LogFactory not found by adding a module dependency
* on apache logging.  This is necessary because commons-logging became modularized
* between 1.2.0 and 1.3.0 but amazon is expecting the non-modular 1.2.0 while
* 1.3.0 is being brought in due to changes in htsjdk.
* See https://logging.apache.org/blog/2023/12/02/apache-common-logging-1.3.0.html for
* more info.
* Fix for https://github.com/igvteam/igv/issues/1598

@jrobinso  Can you test this on your system before we merge?  It seems to fix the problem for me but I want to be sure it's also fixing it for you.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,8 +137,15 @@ dependencies {
             [group: 'software.amazon.awssdk', name: 'cognitoidentity', version: amazonVersion],
             [group: 'software.amazon.awssdk', name: 'sts', version: amazonVersion],
             [group: 'software.amazon.awssdk', name: 's3', version: amazonVersion],
-            [group: 'software.amazon.awssdk', name: 'sso', version: amazonVersion]
+            [group: 'software.amazon.awssdk', name: 'sso', version: amazonVersion],
+
+            // This is a transitive dependency from amazon but due to modularization changes between
+            // 1.20 and 1.3.0 we need to specify it here.
+            // This can be removed when amazon moves from 1.2.0 -> 1.3.0 and updates it's module requirements
+            // See https://logging.apache.org/blog/2023/12/02/apache-common-logging-1.3.0.html
+            [group: 'commons-logging', name: 'commons-logging', version: '1.3.0']
     )
+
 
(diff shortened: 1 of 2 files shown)


brainstorm:
@jrobinso Please can you expedite this fix and cut a new IGV release ASAP? Our team is currently experiencing some issues that seem to point to this.

/cc @ohofmann @victorskl

jrobinso:
We're targetting a release for the end of the month.  In the meantime I suggest you use 2.18.1.   Alternatively you could use the "snapshot" build, in fact it would be useful to know if that build resolves your issues.  If it doesn't then cutting a new release will not either.    You can install it from https://igv.org/doc/desktop/#DownloadSnapshot/

brainstorm:
Thanks Jim, we had to ship a fix at our data portal frontend (outside of IGV) as a workaround so we're good for now, we'll potentially revert that change and test out the new release or test with our dev environment.

brainstorm:
Jim, tested on our dev environment and indeed it seems to fix the issue.

I'll circulate the snapshot release now to our curators for now, but I'd prefer to have a point release for easier tracking of changes and compliance.

jrobinso:
I am building release 2.19.0, which includes this fix now.   However we are
running into holidays here and it might not be released on the website
until Monday.     I can provide links for download when its notarized (for
Macs), which will be sometime tomorrow.

brainstorm:
Thanks Jim, very much appreciated!!!

jrobinso:
We're having trouble with Mac notarization, that won't get resolved before the holidays.  Links below for Windows and the generic command line build

* Windows: https://data.broadinstitute.org/igv/projects/downloads/2.19/IGV_Win_2.19.0-WithJava-installer.exe
* Generic command line:  https://data.broadinstitute.org/igv/projects/downloads/2.19/IGV_2.19.0.zip

PR #1622, opened 2024-11-20
Add READ_ORDER to the ColorBy dropdown in the Preferences window

DevangThakkar:
The `Color alignments by` option in the `Alignment Track Defaults` section in the `Alignments` tab of the Preferences window does not list `READ_ORDER` as an allowed default option but this functionality is available through the right-click menu. This change allows users to set coloring by read order as the default if needed. Based on my understanding of the code, and looking at the functionality for `READ_ORDER` (an option that currently exists only for GroupBy in the Preferences) vs `READ_GROUP` (an option that exists for both GroupBy and ColorBy in the Preferences), no other change should be needed but please correct me if I am wrong.

Cheers,
Devang


jrobinso:
Looks good, thanks.

PR #1660, opened 2025-02-25
Large track hubs

jrobinso:


maximilianh:
JimR got this done without a trackDb cache … interesting.

jrobinso:
@maximilianh I'm not sure I got his done, or done satisfactorily.

maximilianh:
Ha, no problem, I was just impressed that you are getting there! making
these big hubs work took us a long time, we ended up using binary files in
C, a binary dump of the structs, mmaped, as the parsing of the trackDb
files for every click took so long. When we started track hubs, the plan
was never to have 6000+ tracks in them. We thought about allowing indexed
files for storing the config information, but that was excluded. I imagine
in IGV as a standalone app the parsing is a smaller problem, as you parse
only once, not for every click. great to see this.

PR #1732, opened 2025-08-12
Fixing a bug with locally downloaded 2bit references.

lbergelson:
I noticed a problem when trying to download references that are 2bit.   I would get these errors:
```
SEVERE [Aug 12,2025 17:44] [MessageUtils] Error loading genome hg19<br/>no protocol: /home/whaleberg/igv/genomes/hg19/hg19.chrom.sizes
SEVERE [Aug 12,2025 17:44] [MessageUtils] java.net.MalformedURLException: no protocol: /home/whaleberg/igv/genomes/hg19/hg19.chrom.sizes
        at java.base/java.net.URL.<init>(URL.java:772)
        at java.base/java.net.URL.<init>(URL.java:654)
        at java.base/java.net.URL.<init>(URL.java:590)
        at org.igv/org.broad.igv.feature.genome.Genome.<init>(Genome.java:159)
        at org.igv/org.broad.igv.feature.genome.load.JsonGenomeLoader.loadGenome(JsonGenomeLoader.java:27)
        at org.igv/org.broad.igv.feature.genome.GenomeManager.loadGenome(GenomeManager.java:171)
        at org.igv/org.broad.igv.feature.genome.GenomeManager.loadGenomeById(GenomeManager.java:147)
        at org.igv/org.broad.igv.ui.IGV$StartupRunnable.run(IGV.java:1934)
        at org.igv/org.broad.igv.util.LongRunningTask.call(LongRunningTask.java:73)
        at org.igv/org.broad.igv.util.LongRunningTask.call(LongRunningTask.java:43)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)

SEVERE [Aug 12,2025 17:44] [MessageUtils] Error loading genome hg19<br/>no protocol: /home/whaleberg/igv/genomes/hg19/hg19.chrom.sizes
```

It turns out there were some places in Genome that were assuming that the fields from the json that were called urls were actual http urls when for local files they are local file paths.  I added a branch to the checks to handle local files as well.  

I suspect there are still reasonable file types that aren't handled here, like google paths, but now I can at least download files and not get errors.  I would love to more generally convert things to use the HtsPath and the HttpNio connector but that seems like a huge change to the current state of things.

* Auxiliary files for 2bit references were being loaded as HttpUrls which works for remote files but not local ones.
* Fixes to account for local files

### DIFF
--- a/src/main/java/org/broad/igv/feature/genome/Genome.java
+++ b/src/main/java/org/broad/igv/feature/genome/Genome.java
@@ -56,6 +56,7 @@ import org.broad.igv.track.TribbleFeatureSource;
 import org.broad.igv.ucsc.hub.Hub;
 import org.broad.igv.ucsc.hub.HubParser;
 import org.broad.igv.ucsc.twobit.TwoBitSequence;
+import org.broad.igv.util.FileUtils;
 import org.broad.igv.util.HttpUtils;
 import org.broad.igv.util.ResourceLocator;
 import org.broad.igv.util.liftover.Liftover;
@@ -65,8 +66,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * Simple model of a genome.  Keeps an ordered list of Chromosomes, an alias table, and genome position offsets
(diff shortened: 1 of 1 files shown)


jrobinso (commented):
Looks fine, however in the case of a local file there is no need to check the file size (i.e. there is no maximum size).   The motivation behind this check stems from our support for virtually all NCBI assemblies through GenArk.   Many of these are not really assembled and have millions of contigs,  so the chromSizes file is 100s of MBs.   Really this check should probably not be here, it should be in the code that generates a "GenomeConfig" from a Genark stanza,  but here it is.  

So I suggest "fileSizeIsOk" just return true if the file is local.

PR #1764, opened 2025-10-17
Add support for coloring alignments by supplementary flag

sjl:
Hi, I've been using IGV for a while now and have been making small changes to make my daily work easier.  I think some of them might be helpful to other folks too, so I figured I'd send a PR and see if you're interested in pulling them.  If not, no worries.

This patch adds the ability to color alignments by the supplementary flag, which makes it a lot easier to notice them.

Before:

After:


jrobinso:
Thanks for the PR.  I agree this is useful, but if the intent is to highlight split (chimeric) alignments should we also include "representative" alignment, assuming it has an SA tag,  and maybe label it "Split" alignments?  So the idea would be color alignments with either  the supplementary flag or an SA tag.

sjl:
> Thanks, this looks good. I'll get to it when I can.

Thanks!
 
> As I understand the spec there is no meaningful difference between the "primary" and supplemental alignments, one is arbitrarily picked as primary (or "representative").

According to the SAM spec, that's true.  In practice, I think aligners try to choose the best alignment to be the primary alignment (for some value of "best").  For example: minimap2 will make the alignment with the best score the primary alignment, only randomly choosing as a fallback if there are multiple alignments with the same highest score:

https://github.com/lh3/minimap2/issues/323#issuecomment-461034414

https://github.com/lh3/minimap2/issues/605#issuecomment-660142473

jrobinso:
Thanks for the info.  Worth maintaining the distinction then.   I'll do a quick review and merge ASAP,  ping again if this lingers.

jrobinso:
Looks good!  Merged

PR #1792, opened 2025-12-19
Add methylation strandness

armintoepfer:
Studying hemimethylations is tricky, if we can't visualize them nicely. I added two features to address that gap.

# Coloring
Currently, the full rectangle is colored for a methylated base and I can't see which strand it's on:

image

I added the feature to visualize forward + strand methylations by only coloring the top half of the rectangle; lower half corresponds to the rev - strand. For 5hmC is looks like this

image

and for 6mA on the reverse strand only case

image

and squished mode

image

# Popup
Add modified strand to methylation information

image

image

image

<hr>

Cheers to @amwenger for the suggestion and @williamrowell for the special 6mA data

### DIFF
--- a/src/main/java/org/broad/igv/sam/AlignmentTrack.java
+++ b/src/main/java/org/broad/igv/sam/AlignmentTrack.java
@@ -1309,6 +1309,10 @@ public class AlignmentTrack extends AbstractTrack implements IGVEventObserver {
             return baseQualityMax;
         }
 
+        public AlignmentTrack getTrack() {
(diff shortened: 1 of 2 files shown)


jrobinso:
@armintoepfer Did you consider grouping alignments by strand?

jrobinso:
@amwenger @armintoepfer I merged this as it does no harm with the added user preference.  However I don't have a good test dataset, could one of you make the test dataset illustrated available?

PR #1804, opened 2026-01-05
Fix rendering issue by repainting

parambath92:
We noticed some display refresh problems with Swing dialogs where the window doesn't repaint itself when we move away from the IGV tab in browser to some other tab or window in the machine. We have to move the cursor over the page multiple times to get it repainted with delay. Attaching screenshot for reference.
Screenshot 2026-01-05 201438
This PR solves the rendering issues when forcing Swing dialogs (instead of native dialogs). The timer ensures windows are periodically repainted to avoid visual glitches or frozen UI elements.

### DIFF
--- a/src/main/java/org/broad/igv/Globals.java
+++ b/src/main/java/org/broad/igv/Globals.java
@@ -29,7 +29,6 @@ import org.broad.igv.logging.*;
 
 import java.text.DecimalFormat;
 import java.util.*;
-import java.util.List;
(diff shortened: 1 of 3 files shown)


parambath92:
This PR can be closed as its fixed in newer version of xpra-html5 package within IGV. The issue happened because of a bug on xpra-html5 package in IGV as mentioned in the [issue here](https://github.com/Xpra-org/xpra-html5/issues/397).

PR #1835, opened 2026-05-18
Fix base modification filter canonical base lookup

kojix2:
Fix the canonical base passed to the base modification filter

Currently, `canonicalBase` is passed to the filter before it is updated from `bmSet`. As a result, the filter may receive the initial value instead of the actual canonical base.

This change passes `bmSet.getCanonicalBase()` directly when applying the filter.


jrobinso:
@kojix2 thanks for the PRs!

PR #1850, opened 2026-07-29
fix: skip batch mode for HTTP /load to allow parallel track loading

### LINKED ISSUE #1849: Parallel track loading when triggered via HTTP command listener
When loading multiple tracks via the HTTP command listener (port 60151), the `loadFiles` method in `CommandExecutor.java` sets `Globals.setBatch(true)` to force synchronization. This causes all tracks to load sequentially even when parallel loading would be possible.

For use cases where multiple remote BAM files are loaded (e.g. 11 PacBio tracks from S3), this results in total load times of ~25 sec × N tracks. With 11 tracks that's ~5 minutes.

**Example request:**
```
GET /load?merge=false&genome=b37&file=https://example.net/.../BAM1.bam,https://example.net/.../BAM2.bam,...&locus=6:32006087-32009273
```

**Observed behavior:** Tracks load one by one sequentially 

**Expected behavior:** Tracks load in parallel, reducing total time to roughly the time of a single track.

**Root cause:** In `CommandExecutor.java`, `loadFiles` forces batch mode before calling `igv.loadTracks(fileLocators)`:
```java
// Temporarily switch to batch mode to force synchronization of load followed by possible sort
Globals.setBatch(true);
```

**Request:** Allow parallel track loading via the command listener when no sort operation is required, or add an optional `parallel=true` parameter to the `/load` command.

**Reference:** Related to #795 (parallel BAM processing) which was addressed for interactive use but not for the command listener path.

bsgarciac:
## Summary

Fixes the sequential track loading issue reported in #1849.

When loading multiple tracks via the HTTP command listener (`/load`), `loadFiles` was unconditionally calling `Globals.setBatch(true)`, which forced synchronous/sequential loading. This is necessary for port/batch script commands (where the next command is unknown), but not for HTTP requests where all files are known upfront.

**Changes:**
- Added `boolean forceBatch` parameter to `loadFiles` in `CommandExecutor`
- Port/batch script path passes `forceBatch=true` (no behavior change)
- HTTP `/load` path passes `forceBatch=false`, allowing IGV to load tracks without forcing batch mode

## Impact

- **HTTP `/load`**: tracks can now load in parallel instead of sequentially
- **Port/batch scripts**: no change, batch mode still enforced
- **Sort after load**: unaffected, sort is applied after `loadTracks` returns regardless of batch mode

## Test case

Loading 11 remote BAM files via HTTP `/load` previously took ~4.5 minutes (11 × ~25 sec sequential). With this change, loading time should drop significantly as tracks load in parallel.

Closes #1849

### DIFF
--- a/src/main/java/org/igv/batch/CommandExecutor.java
+++ b/src/main/java/org/igv/batch/CommandExecutor.java
@@ -1,4 +1,3 @@
-
 package org.igv.batch;
 
 import org.igv.Globals;
@@ -709,7 +708,7 @@ public class CommandExecutor {
         Map<String, String> ignore = null;
         String sort = null;
         String sortTag = null;
-        return loadFiles(fileString, index, coverage, name, format, locus, merge, ignore, sort, sortTag, true);
+        return loadFiles(fileString, index, coverage, name, format, locus, merge, ignore, sort, sortTag, true, true);
 
     }
 
@@ -724,6 +723,8 @@ public class CommandExecutor {
      * @param sort
      * @param sortTag    Used iff sort == SortOption.TAG
      * @param dup
+     * @param forceBatch if true, enables batch mode to synchronize loading (required for port/batch scripts);
+     *                   pass false for HTTP /load requests to allow parallel track loading
      * @return
      * @throws IOException
      */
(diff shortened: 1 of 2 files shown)


jrobinso:
This looks good.

bsgarciac:
Thanks @jrobinso 
quick question, when is the next beta version expected to be released?

